PR #6, opened 2016-01-22
global: refactoring of models

lnielsen:
* Refactors models to support sharing the same files between multiple buckets.

Signed-off-by: Lars Holm Nielsen <lars.holm.nielsen@cern.ch>

### DIFF
--- a/invenio_files_rest/models.py
+++ b/invenio_files_rest/models.py
@@ -26,17 +26,21 @@
 
 from __future__ import absolute_import, print_function
 
+import re
 import uuid
 
-from flask import url_for
+from flask import current_app
 from invenio_db import db
-from sqlalchemy import desc
+from sqlalchemy.orm import validates
 from sqlalchemy_utils.models import Timestamp
 from sqlalchemy_utils.types import UUIDType
 
+from .errors import FileInstanceAlreadySetError
+
+slug_pattern = re.compile("^[a-z][a-z0-9-]+$")
+
 
 class Location(db.Model, Timestamp):
-
     """Model defining base locations."""
 
     __tablename__ = 'files_location'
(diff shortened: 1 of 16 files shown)


PXke:
Looks good

invenio_files_rest/config.py
```diff
@@ -22,12 +22,21 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
-"""Invenio File Rest module configuration file."""
+"""Invenio Files Rest module configuration file."""
 
 FILES_REST_STORAGE_CLASS_LIST = {
```
jirikuncar: IMHO it is not configuration but it should be defined as enum in models. Can admin define more storage classes? If yes, then we should have something like: `'S': 'invenio_files_rest.storage:Standard'`.

invenio_files_rest/models.py
```diff
@@ -157,92 +249,120 @@ class Object(db.Model, Timestamp):
...
-                cls.bucket_id == bucket_id,
-                ~cls.deleted,
-                cls.updated == db.session.query(
-                    db.func.max(table_object.updated)).filter(
-                        ~table_object.deleted,
-                        table_object.filename == cls.filename).limit(1)
-            )
+        args = [
+            cls.bucket_id == bucket_id,
+            cls.file_id != None,  # noqa
```
jirikuncar: `cls.file_id.isnot(None),`

invenio_files_rest/config.py
lnielsen: The idea behind storage class is to allow a system to record what type of storage a given file resides on - e.g. offline/online, reduced resilience, ... . In Files-REST there's no specific functionality associated with it, but an Invenio instance may use it to adapt behaviour depending no the storage class.

invenio_files_rest/models.py
lnielsen: Thanks

invenio_files_rest/views.py
```diff
@@ -490,15 +412,91 @@ def get(self, bucket_id, filename, **kwargs):
...
+                {
+                    "checksum": "xxxx",
+                    "size": 0,
+                    "url": "http://localhost:5000/files/322ea4c6-665.../f.pdf",
+                    "uuid": "322ea4c6-6650-4143-a328-274eee55f45d",
+                    "updated": "2015-12-10T14:16:57.202795"
+                }
+
+            :resheader Content-Type: application/json
+            :statuscode 200: no error
```
blixhavn: Shouldn't this be 201/202?

tests/conftest.py
```diff
@@ -48,53 +48,45 @@ def app(request):
...
-    Babel(app_)
-    InvenioFilesREST(app_)
-
-    app_.config['SQLALCHEMY_DATABASE_URI'] = os.environ.get(
-        'SQLALCHEMY_DATABASE_URI', 'sqlite:///test.db'
+        TESTING=True,
+        SQLALCHEMY_TRACK_MODIFICATIONS=True,
+        SQLALCHEMY_DATABASE_URI=os.environ.get(
+            'SQLALCHEMY_DATABASE_URI',
+            'postgresql+psycopg2://localhost:5432/files'),
```
ddaze: Not sqlite as default?

PR #28, opened 2016-02-26
Basic file serving

ddaze:

PR #29, opened 2016-02-28
tasks: background fixity checker

lnielsen:


lnielsen:
Depends on inveniosoftware/invenio-admin#17

drjova:
@lnielsen from models ``serialize`` is missing. Are we gonna use the ``serializer``s from https://github.com/inveniosoftware/invenio-files-rest/pull/28?

invenio_files_rest/tasks.py
```diff
@@ -0,0 +1,54 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015, 2016 CERN.
```
drjova: only 2016?

invenio_files_rest/admin.py
```diff
@@ -40,33 +47,35 @@ def require_slug(form, field):
...
+    )
+    column_details_list = (
+        'name', 'uri', 'default', 'created', 'updated', 'buckets')
+    column_list = ('name', 'uri', 'default', 'created', 'updated', 'buckets')
     column_labels = dict(
         id=_('ID'),
         uri=_('URI'),
     )
-    column_filters = ('default', 'created', 'updated', )
+    column_filters = ('default', 'created', 'updated')
```
drjova: shall we keep the comma?

lnielsen:
@drjova Serialisers will be rewritten using the same logic as what has just been implemented in Records-REST.

drjova:
@lnielsen ok! the only problem is that the example's API is broken :/ Is it possible to add dummy `serialisers` for now WDYT? Otherwise is good to go :)

lnielsen:
@drjova Yes, #28 is adding some dummy serialisers so we can merge that one right after.

drjova:
:shipit:

PR #31, opened 2016-03-02
models: bucket tags

lnielsen:


lnielsen:
Tags as discussed IRL

egabancho:
LGTM, @krzysztof?

examples/app.py
```diff
@@ -133,22 +142,22 @@ def files():
     # Bucket 1
     b1 = Bucket.create(loc)
     for f in ['README.rst', 'LICENSE']:
-        with open(join(srcroot, f), 'r') as fp:
-            Object.create(b1, f, stream=fp)
+        with open(join(srcroot, f), 'rb') as fp:
```
krzysztof: Not a real issue but for testing purposes I find ``BytesIO`` object to be a cleaner solution, see for example: https://github.com/krzysztof/invenio-files-rest/blob/permissions/tests/test_permissions.py#L40

krzysztof:
Well, I guess it LGTM as well then ;)

PR #32, opened 2016-03-02
examples: import correction

drjova:
Signed-off-by: Harris Tzovanakis <me@drjova.com>


lnielsen:
This is already fixed in #29 - just needs to be integrated. Can you perhaps review and merge #29 ?

PR #46, opened 2016-03-30
global: new bucket quota validation

### LINKED ISSUE #25: quotas: simple support
Add support for simple quotas - e.g. quota per bucket.

nharraud:
* NEW Adds bucket quota validation when files are added (closes #25).

Signed-off-by: Nicolas Harraudeau <nicolas.harraudeau@cern.ch>

### DIFF
--- a/invenio_files_rest/models.py
+++ b/invenio_files_rest/models.py
@@ -193,9 +193,11 @@ class Bucket(db.Model, Timestamp):
     """
 
     quota_size = db.Column(db.BigInteger, nullable=True)
-    """Quota size of bucket.
+    """Quota size of bucket. Used only by the file_size_limiter.
 
-    A ``None`` value means no quota.
+    Note, don't use this attribute directly. It MAY be used to store
+    the actual quota size for this bucket. Change the file_size_limiter if
+    you want to filter accepted files based on their size.
     """
 
     locked = db.Column(db.Boolean, default=False, nullable=False)
@@ -628,7 +630,10 @@ class ObjectVersion(db.Model, Timestamp):
         """Set a file instance."""
         if self.file_id is not None:
             raise FileInstanceAlreadySetError()
+
         self.file = fileinstance
+        self.bucket.size += self.file.size
+
         return self
 
     def restore(self):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -73,6 +73,7 @@ def base_app():
         SECURITY_PASSWORD_SALT='TEST_SECURITY_PASSWORD_SALT',
         SECRET_KEY='TEST_SECRET_KEY',
     )
+
     FlaskCLI(app_)
     FlaskCeleryExt(app_)
     InvenioDB(app_)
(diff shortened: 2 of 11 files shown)


examples/app.py
```diff
@@ -78,11 +78,11 @@
 from invenio_admin import InvenioAdmin
 from invenio_celery import InvenioCelery
 from invenio_db import InvenioDB, db
-from invenio_rest import InvenioREST
```
lnielsen: I think this one should stay as is.

invenio_files_rest/models.py
```diff
@@ -192,10 +193,13 @@ class Bucket(db.Model, Timestamp):
     inside the bucket.
     """
 
-    quota_size = db.Column(db.BigInteger, nullable=True)
-    """Quota size of bucket.
+    internal_quota = db.Column(db.BigInteger, nullable=True)
```
lnielsen: Can we call it ``quota`` instead.

invenio_files_rest/models.py
```diff
@@ -575,7 +585,7 @@ def is_deleted(self):
         return self.file_id is None
 
     def set_contents(self, stream, size=None, chunk_size=None,
-                     progress_callback=None):
+                     progress_callback=None, expected_size=None):
```
lnielsen: ``expected_size`` is unused. Also I think ``size`` equals the expected size?

invenio_files_rest/models.py
```diff
@@ -624,11 +643,23 @@ def set_location(self, uri, size, checksum, storage_class=None):
 
         return self
 
-    def set_file(self, fileinstance):
+    def set_file(self, fileinstance, ignore_quota=False):
```
lnielsen: Should the quota also be checked in ``set_location``?

tests/conftest.py
```diff
@@ -73,6 +73,10 @@ def base_app():
         SECURITY_PASSWORD_SALT='TEST_SECURITY_PASSWORD_SALT',
         SECRET_KEY='TEST_SECRET_KEY',
     )
+    # update the application with the configuration provided by the test
```
lnielsen: I would prefer not to use the ``@pytest.mark.parametrize`` as it IMHO makes the tests much harder to read and understand.

tests/conftest.py
```diff
@@ -82,8 +86,8 @@ def base_app():
     return app_
 
 
-@pytest.yield_fixture(scope='session', autouse=True)
-def app(base_app):
+@pytest.yield_fixture(scope='function', autouse=True)
```
lnielsen: I assume this is because of ``@pytest.mark.parametrize``?

tests/conftest.py
```diff
@@ -153,10 +157,16 @@ def bucket(db, dummy_location):
 @pytest.yield_fixture()
 def objects(db, bucket):
     """File system location."""
+    data_bytes = b('license file')
     obj1 = ObjectVersion.create(
-        bucket, 'LICENSE', stream=BytesIO(b('license file')))
+        bucket, 'LICENSE', stream=BytesIO(),
```
lnielsen: ``BytesIO(data_bytes)``?

tests/test_views_records_ui.py
```diff
@@ -32,9 +32,10 @@
 from flask import url_for
 from invenio_pidstore import InvenioPIDStore
 from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+from invenio_records_ui import InvenioRecordsUI
+
```
lnielsen: Check import sorting.

lnielsen:
Remember to add yourself to AUTHORS as well.

tests/conftest.py
nharraud: In this case I am not sure that the tests are bigger. Only the tests needing a specific application configuration are parametrized. As we set the configuration before initializing the extensions, we would need to move all this init in each test, thus making them even bigger and duplicating code.
It also enables easy testing with different configurations.
However it's true that the readability is far from being perfect.

Or we can also suppose that the options can be set after init time (in this case it would work), but I would prefer to test the code the same way we recommend to use it.

invenio_files_rest/models.py
nharraud: Yes, removing it right away.

lnielsen:
LGTM. Please rebase and I'll merge.

nharraud:
@lnielsen ready

PR #48, opened 2016-04-29
global: object key and file uri db type change

krzysztof:
* Changes the type of file URI and object key to db.Text with validation.

Signed-off-by: Krzysztof Nowak <k.nowak@cern.ch>


krzysztof:
It seems that MySQL cannot have TEXT as primary key...

PR #51, opened 2016-05-02
bug fix: rename filename to key in DELETE and HEAD http calls

### LINKED ISSUE #50: views: bad parameter name
Calling HTTP DELETE (and probably also HEAD) on an ObjectResource fails with `TypeError: delete() missing 1 required positional argument: 'filename'`. The `filename` argument should be renamed to `key`.

https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/views.py#L614
https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/views.py#L667

emanueldima:
Signed-off-by: Emanuel Dima <emanueldima@gmail.com>

Closes #50

### DIFF
--- a/invenio_files_rest/views.py
+++ b/invenio_files_rest/views.py
@@ -611,7 +611,7 @@ class ObjectResource(ContentNegotiatedMethodView):
             db.session.rollback()
             abort(400, 'File size different than Content-Length')
 
-    def delete(self, bucket_id, filename, **kwargs):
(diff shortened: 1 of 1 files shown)


lnielsen:
Thanks for the fix. Would you mind to change the commit message to:

```
views: filename to key renaming

Signed-off-by: Emanuel Dima <emanueldima@gmail.com>
```

PR #55, opened 2016-05-19
cli: import from directory

jirikuncar:


invenio_files_rest/cli.py
```diff
@@ -0,0 +1,104 @@
...
+
+@files.command()
+@click.argument('name')
+@click.argument('uri')
+@click.option('--default', is_flag=True, default=False)
+@with_appcontext
+def location(name, uri, default):
+    """Create new location."""
+    from .models import Location
+    location = Location(name=name, uri=uri, default=default)
```
lnielsen: The previous default location should be unset. Perhaps implemented via a ``Location.create()``?

invenio_files_rest/cli.py
```diff
@@ -0,0 +1,104 @@
...
+    db.session.commit()
+    click.secho(str(bucket), fg='green')
+
+
+@files.command()
+@click.argument('name')
+@click.argument('uri')
+@click.option('--default', is_flag=True, default=False)
+@with_appcontext
+def location(name, uri, default):
```
lnielsen: Would it make sense to also allow modifying the uri of an existing location using the same cli command?

invenio_files_rest/cli.py
```diff
@@ -0,0 +1,104 @@
...
+@click.group()
+def files():
+    """File management commands."""
+
+
+@files.command()
+@click.argument('source', type=click.Path(exists=True, resolve_path=True))
+@click.option('--checksum/--no-checksum', default=False)
+@with_appcontext
+def bucket_from(source, checksum):
```
lnielsen: Would it perhaps make sense to split it in two commands into:

- ``bucket`` (create a bucket)
- ``cp <source> <bucket, key?>`` (then we could later extend with ``cp <bucket, key?> <destination>`` as well as ``cp <src bucket, key> <dest bucket, key>``) so we have basic upload/download cli.

I.e. basically imitating unix commands.

invenio_files_rest/cli.py
jirikuncar: We can create an issue if somebody needs it otherwise I can remove the cli completely and keep only the helper function.

PR #61, opened 2016-05-24
models: per bucket file size limits

SamiHiltunen:
* Adds a per bucket maximum file size limit.

* File size limits are now checked while writing the object.

### DIFF
--- a/invenio_files_rest/config.py
+++ b/invenio_files_rest/config.py
@@ -38,10 +38,20 @@ the file and/or what is the reliability.
 FILES_REST_DEFAULT_STORAGE_CLASS = 'S'
 """Default storage class."""
 
-FILES_REST_STORAGE_FACTORY = None
+FILES_REST_DEFAULT_QUOTA_SIZE = None
+"""Default quota size for a bucket in bytes."""
+
+FILES_REST_DEFAULT_MAX_FILE_SIZE = None
+"""Default maximum file size for a bucket in bytes."""
+
+FILES_REST_SIZE_LIMITERS = 'invenio_files_rest.limiters.file_size_limiters'
+"""Import path of file size limiters factory."""
+
+FILES_REST_STORAGE_FACTORY = 'invenio_files_rest.storage.pyfs_storage_factory'
 """Import path of factory used to create a storage instance."""
 
-FILES_REST_PERMISSION_FACTORY = None
+FILES_REST_PERMISSION_FACTORY = \
+    'invenio_files_rest.permissions.permission_factory'
 """Import path of permission factory."""
 
 FILES_REST_OBJECT_KEY_MAX_LEN = 255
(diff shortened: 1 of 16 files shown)


invenio_files_rest/config.py
```diff
@@ -38,6 +38,15 @@
 FILES_REST_DEFAULT_STORAGE_CLASS = 'S'
 """Default storage class."""
 
+FILES_REST_DEFAULT_QUOTA_SIZE = None
+"""Default quota size for a bucket."""
+
+FILES_REST_DEFAULT_MAX_FILE_SIZE = None
+"""Default maximum file size for a bucket."""
+
+FILES_REST_SIZE_LIMITERS = None
```
lnielsen: ``/FILES_REST_SIZE_LIMITERS/FILES_REST_SIZE_LIMITER/`` plural -> singular ?

lnielsen: or alternatively the doc text should be in plural

invenio_files_rest/limiters.py
```diff
@@ -0,0 +1,84 @@
...
+            'Maximum file size exceeded.',
+        ),
+    ]
+
+
+class FileSizeLimit(object):
+    """File size limiter."""
+
+    not_implemented_error = NotImplementedError(
+        'FileSizeLimit support only comparisons withints and other '
```
lnielsen: ``/withints/with integers/``

invenio_files_rest/views.py
```diff
@@ -595,21 +582,22 @@ def put(self, bucket_id, key, content_length=None, content_md5=None):
...
-                'verisionId': str(obj.version_id),
-            }}
+            return {
+                'json': {
+                    'checksum': obj.file.checksum,
+                    'size': obj.file.size,
+                    'versionId': str(obj.version_id),
+                }
+            }
+        except FileSizeError as e:
```
lnielsen: If we make ``FileSizeError`` a subclass of ``invenio_rest.errors:RESTException`` and wrap above block in a ``with db.session.begin_nested():``, then we can probably completely remove the try/except block (i.e. also the SQLAlchemyError, as it will properly rollback and return a 500 error.

invenio_files_rest/views.py
```diff
@@ -583,36 +576,24 @@ def put(self, bucket_id, key, content_length=None, content_md5=None):
...
         # TODO: Support setting content-type
         # TODO: Don't create a new file if content is identical.
 
-        try:
-            # TODO: Pass storage class to get_or_create
+        # TODO: Pass storage class to get_or_create
+        with db.session.begin_nested():
             obj = ObjectVersion.create(bucket, key)
-            obj.set_contents(uploaded_file, size=content_length)
-            db.session.commit()
```
lnielsen: Don't you still need a ``db.session.commit()`` outside the ``with``-statement?

SamiHiltunen:
@lnielsen I've fixed the issues you found

PR #66, opened 2016-05-27
models: snapshot size population

SamiHiltunen:
* FIX Sets the size for Bucket snapshots.


invenio_files_rest/models.py
```diff
@@ -249,6 +249,7 @@ def snapshot(self, lock=False):
                 default_location=self.default_location,
                 default_storage_class=self.default_storage_class,
                 quota_size=self.quota_size,
+                size=self.size,
```
lnielsen: This would not be correct, since we're only copying the latest head ObjectVersion and not any previous versions. I've already fixed the problem in #65

PR #67, opened 2016-05-27
views: BucketCollectionView serializers

SamiHiltunen:
* Adds a serializer for BucketCollectionView.


invenio_files_rest/models.py
```diff
@@ -298,12 +299,43 @@ def get(cls, bucket_id):
...
+        """Delete a bucket.
+
+        Does not actually delete the Bucket, just marks it as deleted.
+        Deleting a bucket also marks the contained objects as deleted.
+        """
+        bucket = cls.get(bucket_id)
+        if not bucket or bucket.deleted:
+            return False
+
+        for obj in ObjectVersion.query.filter_by(bucket_id=bucket_id):
```
lnielsen: I'm not sure if we should mark the objects as deleted. Say you want to undelete an entire bucket.

invenio_files_rest/serializer.py
```diff
@@ -50,3 +51,45 @@ def json_serializer(data=None, code=200, headers=None):
...
+    def decorator(data=None, code=200, headers=None):
+        data = data if not data else f(data, code, headers)
+        return json_serializer({'json': data}, code, headers)
+    return decorator
+
+
+@empty_if_none
+def bucket_collection_serializer(data=None, code=200, headers=None):
+    """Serialize BucketCollectionResource responses."""
+    def serialize(bucket):
```
lnielsen: We should try to match as close as possible the S3 REST API: http://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html

invenio_files_rest/views.py
```diff
@@ -60,7 +128,7 @@ def file_download_ui(pid, record, **kwargs):
         RECORDS_UI_ENDPOINTS = dict(
             recid=dict(
                 # ...
-                route='/records/<pid_value/files/<filename>',
```
lnielsen: In this case I actually think it's correct with ``filename``.

invenio_files_rest/views.py
```diff
@@ -92,20 +154,22 @@ def file_download_ui(pid, record, **kwargs):
...
             **kwargs
         )
-        self.post_args = {
-            'location_name': fields.String(
-                missing=None,
-                location='json'
-            )
-        }
 
+    @need_buckets_permission('buckets-read')
```
lnielsen: Should by default be limited to admins only - or no one - whatever is easier.

invenio_files_rest/views.py
```diff
@@ -149,19 +213,12 @@ def get(self, **kwargs):
...
-            })
-        # FIXME: how to avoid returning a dict with key 'json'
-        return {'json': bucket_list}
-
-    def post(self, **kwargs):
+        # TODO: Only get buckets that user has permission to.
+        return Bucket.all()
+
+    @use_kwargs(post_args)
+    @need_buckets_permission('buckets-create')
```
lnielsen: Same here - admins only.

invenio_files_rest/views.py
```diff
@@ -205,56 +262,40 @@ def post(self, **kwargs):
...
-            'versions': fields.Boolean(
-                location='query'
-            )
-        }
 
-    def get(self, bucket_id, **kwargs):
+    @use_kwargs(get_args)
+    @pass_bucket
+    @need_bucket_permission('bucket-read')
+    def get(self, versions, bucket=None, **kwargs):
```
lnielsen: We need to be able to probably distinguish between allowing list of versions or not.
http://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html

invenio_files_rest/views.py
```diff
@@ -411,7 +433,7 @@ def head(self, bucket_id=None, **kwargs):
             :statuscode 403: access denied
             :statuscode 404: the bucket does not exist
         """
-        if not bucket_id or not Bucket.get(bucket_id):
+        if not bucket:
```
lnielsen: Isn't this already checked by ``@pass_bucket``

lnielsen:
@SamiHiltunen Will you push the latest code?

invenio_files_rest/views.py
```diff
@@ -50,6 +55,92 @@
...
+    @wraps(f)
+    def decorate(*args, **kwargs):
+        uploaded_file = request.files.get('file')
+        if not uploaded_file:
+            abort(400, 'File is missing from the request.')
+        return f(uploaded_file=uploaded_file, *args, **kwargs)
+    return decorate
+
+
+def need_bucket_collection_permission(action):
```
lnielsen: Perhaps something along the lines of to avoid duplication of the needs_permission logic:

```python
def need_permissions(factory, action):
    """"Get permission for buckets or abort."""
    def decorator_builder(f):
        @wraps(f)
        def decorate(*args, **kwargs):
            permission = factory(kwargs, action)
            if permission is not None and not permission.can():
                if current_user.is_authenticated:
                    abort(403,
                          'You do not have a permission for this action')
                abort(401)
            return f(*args, **kwargs)
        return decorate
    return decorator_builder


need_bucket_collection_permission = partial(
    need_permission, 
    lambda kwargs, action: current_bucket_collection_permission_factory(action)
)

need_bucket_permission = partial(
    need_permission, 
    lambda kwargs, action: current_bucket_permission_factory(kwargs['bucket'], action=action)
)

need_objects_permission = partial(
    need_permission, 
    lambda kwargs, action: current_bucket_collection_permission_factory(kwargs['bucket'], kwargs['object'], action=action)
)
```

invenio_files_rest/views.py
```diff
@@ -149,19 +236,12 @@ def get(self, **kwargs):
...
-                'url': url_for("invenio_files_rest.bucket_api",
-                               bucket_id=bucket.id, _external=True),
-                'uuid': str(bucket.id)
-            })
-        # FIXME: how to avoid returning a dict with key 'json'
-        return {'json': bucket_list}
-
-    def post(self, **kwargs):
+        # TODO: Only get buckets that user has permission to.
+        return Bucket.all()
```
lnielsen: By default this method is restricted right?

invenio_files_rest/views.py
```diff
@@ -205,56 +285,40 @@ def post(self, **kwargs):
...
-
-        # TODO: Implement serializer
-        return {'json':
-                {'size': bucket.size,
-                 'url': url_for("invenio_files_rest.bucket_api",
-                                bucket_id=bucket.id, _external=True),
-                 'uuid': str(bucket.id)
-                 }
-                }
+        db.session.commit()
```
lnielsen: Do we have error handlers registered, so that we don't get a 500 error if e.g. the user specifies an invalid location?

PR #68, opened 2016-05-31
storage: addition of hierarchical directory

### LINKED ISSUE #64: storage: hierarchical directory structure

JavierDelgadoFernandez:
* Adds the possiblity to store the files grouped by file name in
  different directories. (closes #64)

Signed-off-by: Javier Delgado <javier.delgado.fernandez@cern.ch>

### DIFF
--- a/invenio_files_rest/storage.py
+++ b/invenio_files_rest/storage.py
@@ -31,6 +31,7 @@ import time
 from functools import partial
 from os.path import join
 
+from flask import current_app
 from fs.opener import opener
 
 from .errors import FileSizeError, StorageError, UnexpectedFileSizeError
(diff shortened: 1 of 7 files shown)


invenio_files_rest/storage.py
```diff
@@ -158,10 +160,18 @@ def __init__(self, fileinstance, base_uri=None, filename='data'):
         self.filename = filename
         super(PyFilesystemStorage, self).__init__(fileinstance)
 
-    def make_path(self):
+    def make_path(self, path_dimensions=FILES_REST_STORAGE_PATH_DIMENSIONS,
```
jirikuncar: please use following pattern it respects the application config:

```python
def foo(bar=None):
    bar = bar or current_app.config['BAR_DEFAULT']
```

tests/test_storage.py
```diff
@@ -64,10 +64,6 @@ def test_file_save(data, **kwargs):
         assert "md5:{0}".format(m.hexdigest()) == checksum
 
         assert size == len(data)
-        assert loc == join(
```
jirikuncar: why did you remove this test?

tests/test_tasks.py
```diff
@@ -58,7 +58,6 @@ def test_migrate_file(app, db, dummy_location, extra_location, bucket,
 
     # Test pre-condition
     old_uri = obj.file.uri
-    assert join(bucket.location.uri, str(obj.file.id), 'data') == old_uri
```
jirikuncar: ! https://github.com/inveniosoftware/invenio-files-rest/pull/68/files#r65142159

tests/test_storage.py
JavierDelgadoFernandez: Now there is a dedicated test checking if the paths are generated correctly. I think it is not good to keep this checking here because, moreover, it should be adapted to use the new path creations, and that requires the develop of the same method which generates the paths here again... If you hardcode the expected path for testing here, it will crash if the user modify the configuration to something else.

lnielsen:
```python
join(dummy_location.uri, str(obj.file.id)[0:2], str(obj.file.id)[:2], "data")
```

tests/test_storage.py
```diff
@@ -161,3 +157,31 @@ def test_storage_interface():
     pytest.raises(NotImplementedError, s.send_file)
     pytest.raises(NotImplementedError, s.save, None)
     pytest.raises(NotImplementedError, s.compute_checksum, None)
+
+
+def test_pyfilesystemstorage_make_path():
+    """Test path for files."""
+    f = PyFilesystemStorage(
+        FileInstance(uri="Test"), base_uri='N')
```
lnielsen: ``python
f = FileInstance.create(...)
f.id = UUID('deadbeef-....')
```

JavierDelgadoFernandez:
@lnielsen Updated.

PR #81, opened 2016-06-29
views: ng-file-upload support fix

lnielsen:
Ping @drjova

PR #86, opened 2016-07-14
views: fix of size check in multipart upload 

drjova:


jirikuncar:
@drjova [fix](https://encrypted.google.com/search?q=define:fix) or [fixation](https://encrypted.google.com/search?q=define:fixation)?

lnielsen:
@drjova I've updated #81 with this commit, fixed the build error and fixed another permission issue that I introduced in the other mega commit.

PR #91, opened 2016-07-29
models: bucket removal addition

lnielsen:


invenio_files_rest/models.py
```diff
@@ -433,9 +433,16 @@ def delete(cls, bucket_id):
 
     @ensure_unlocked()
     def remove(self):
-        """Permanently remove a bucket and all objects."""
-        # 1) remove all object versions 2) remove bucket.
-        raise NotImplementedError()
+        """Permanently remove a bucket and all objects (including versions).
+
+        Note the method does not remove the associated file instances which
```
jirikuncar: Shall we enhance the note with `.. note::`?

lnielsen: Not necessary IMHO as it's doesn't need a framed box around it in the rendered documentation.

slint:
:shipit:

invenio_files_rest/models.py
jirikuncar: There is one in `BucketTag.remove` method.

lnielsen: ``BucketTag.remove`` -> ``ObjectVersion.remove()``?

With:

> Shall we enhance the note with .. note::?

Did you mean:

1)
```
.. note:: 

   Note the method does not remove the associated file instances which 
   must be garbage collected.
```

or:

2)
```
.. note::

   This should only be used when you want to permanently remove a bucket.
```

jirikuncar: Kind of ...

lnielsen: ?

jirikuncar: Both joined as one `.. note::`? I was mostly wondering why one remove method need framed box and the other doesn't when both messages are quite important.

lnielsen: Amended in #92

PR #92, opened 2016-08-01
docs: remove method warning

lnielsen:


jirikuncar:
@lnielsen `docs: remove method warning` sounds a bit misleading as if we were removing method warnings.

lnielsen:
Extra commit added.

invenio_files_rest/models.py
```diff
@@ -377,7 +377,7 @@ def get_tags(self):
         return {t.key: t.value for t in self.tags}
 
     @classmethod
-    def create(cls, location=None, storage_class=None):
+    def create(cls, location=None, storage_class=None, **kwargs):
```
jirikuncar: @lnielsen can you include new arguments in docstring?
```
:param \**kwargs: Keyword arguments are forwarded to the class constructor.
```

jirikuncar:
@lnielsen can you separate the version bump commit to new PR while we are improving this PR?

PR #97, opened 2016-08-09
global: serializer and REST API minor fixes

lnielsen:


invenio_files_rest/views.py
```diff
@@ -590,9 +594,12 @@ def multipart_delete(self, multipart):
     #
     @use_kwargs(get_args)
     @pass_bucket
-    def get(self, bucket=None, key=None, version_id=None, upload_id=None):
+    def get(self, bucket=None, key=None, version_id=None, upload_id=None,
+            uploads=None):
         """Get object or list parts of a multpart upload."""
-        if upload_id:
+        if uploads is not False:
```
jirikuncar: Can we  move it to `get_args` as a `uploads` validator?

PR #100, opened 2016-08-09
view: multipart/form-data support for put

lnielsen:


invenio_files_rest/ext.py
```diff
@@ -92,6 +93,14 @@ def part_factories(self):
             self.app.config.get('FILES_REST_MULTIPART_PART_FACTORIES', [])
         ]
 
+    @cached_property
+    def upload_factories(self):
+        """Factory getting list of webargs schemas for parsing part number."""
+        return [
+            import_string(x) for x in
```
jirikuncar: It would be nice to support callables too.

lnielsen: I will open separate issue for it, as it's all of the methods in ``ext.py`` that has this problem.

lnielsen: #101

jirikuncar: Ok. We can merge it as is.

lnielsen:
@drjova This PR implements the two things we discussed. I've tested on Zenodo and it now works to upload small files, and won't return for POST-request to complete multipart upload until task has been run (or timeout reached).

lnielsen:
Includes #97

lnielsen:
Ping @nharraud - Please note the change of REST API when completing a multipart upload.

invenio_files_rest/config.py
```diff
@@ -89,6 +89,13 @@
 ]
 """Import path of factory used to parse chunked upload parameters."""
 
+FILES_REST_UPLOAD_FACTORIES = [
+
```
jirikuncar: can we remove the empty line?

invenio_files_rest/config.py
```diff
@@ -100,3 +107,9 @@
 
 FILES_REST_MULTIPART_EXPIRES = timedelta(days=4)
 """Time delta after which a multipart upload is considered expired."""
+
+FILES_REST_TASK_WAIT_INTERVAL = 2.0
```
jirikuncar: Why do we use `timedelta` for `FILES_REST_MULTIPART_EXPIRES` and seconds for `FILES_REST_TASK_WAIT_INTERVAL` and `FILES_REST_TASK_MAX_WAIT`? Can we unify the units or at least add the unit in docstring?

invenio_files_rest/views.py
```diff
@@ -123,6 +148,32 @@ def ngfileupload_partfactory(part_number=None, content_length=None,
...
+    ),
+    'content_type': fields.Str(
+        load_from='Content-Type',
+        location='headers',
+        required=True,
+    ),
+})
+def ngfileupload_uploadfactory(total_size=None, content_type=None):
+    """Get default put factory."""
+    if not content_type.startswith('multipart/form-data'):
```
jirikuncar: `if content_type is None or not content_type.startswith ...`

invenio_files_rest/config.py
```diff
@@ -100,3 +107,9 @@
...
+"""Interval between sending a whitespace to prevent connection closing."""
+
+FILES_REST_TASK_MAX_WAIT = 600
```
jirikuncar: `FILES_REST_TASK_MAX_WAIT_TIME`?

invenio_files_rest/config.py
lnielsen: timedelta vs seconds: because of the time scales they are operating on (days, months, years) for multipart expiry vs seconds for the task waiting.

invenio_files_rest/config.py
lnielsen: ``FILES_REST_TASK_WAIT_MAX_SECONDS``

invenio_files_rest/config.py
jirikuncar: I was just curious and looking for uniformity.

WDYT about:
```python
rounds = float(timedelta(minutes=5).total_seconds()) // timedelta(seconds=2).total_seconds()
```

lnielsen: I also like uniformity, but in this context I don't think it makes sense to use ``timedelta``. I associate ``timedelta``  with in date/datetime calculations, not with short timeouts.

jirikuncar: @lnielsen ok. Then let's just change 2.0 to 2.

drjova:
:shipit:

PR #125, opened 2016-10-04
global: fix the last modified conversion

krzysztof:
* Fixes the incorrect conversion of last modified timestamps
  from localtime to UTC.

Signed-off-by: Krzysztof Nowak <k.nowak@cern.ch>

### DIFF
--- a/invenio_files_rest/storage/base.py
+++ b/invenio_files_rest/storage/base.py
@@ -28,6 +28,7 @@ from __future__ import absolute_import, print_function
 
 import hashlib
 import time
+from calendar import timegm
 from functools import partial
 
 from ..errors import FileSizeError, StorageError, UnexpectedFileSizeError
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -75,6 +75,7 @@
     extras_require['all'].extend(reqs)
 
 install_requires = [
+    'arrow>=0.8.0',
```
jirikuncar: Are we sure that we want to introduce new dependency?

krzysztof: alternative is to ``import calendar`` and use https://docs.python.org/2/library/calendar.html#calendar.timegm, but arrow seemed to be more abundant across other packages, so I'd vote yes for the consistency.

jirikuncar: Can we go with `calendar` if it works for you?

jirikuncar:
@krzysztof can you also add test case which would fail without your fix? Thanks

setup.py
krzysztof: :+1:

invenio_files_rest/storage/base.py
```diff
@@ -28,8 +28,10 @@
 
 import hashlib
 import time
+from calendar import timegm
 from functools import partial
 
+
```
jirikuncar: ?

jirikuncar:
LGTM

invenio_files_rest/storage/base.py
krzysztof: Fixed the isort and the unit test, take a look.

jirikuncar:
LGTM

PR #132, opened 2016-10-27

egabancho:
- Introduces new model to store tags (key,value pairs) associated with
  a given object version.

### DIFF
--- a/invenio_files_rest/models.py
+++ b/invenio_files_rest/models.py
@@ -114,6 +114,27 @@ def as_bucket_id(value):
     return value.id if isinstance(value, Bucket) else value
 
 
+def as_object_version(value):
+    """Get a object version object from a object version id or object version.
+
+    :param value: A :class:`invenio_files_rest.models.ObjectVersion` or an
+        object version ID.
+    :returns: A :class:`invenio_files_rest.models.ObjectVersion` instance.
+    """
+    return value if isinstance(value, ObjectVersion) \
+        else ObjectVersion.query.filter_by(version_id=value).one_or_none()
+
+
+def as_object_version_id(value):
+    """Get a object version id from a object version id or object version.
+
+    :param value: A :class:`invenio_files_rest.models.ObjectVersion` instance
+        of a object version id.
+    :returns: The :class:`invenio_files_rest.models.ObjectVersion` version_id.
+    """
+    return value.version_id if isinstance(value, ObjectVersion) else value
+
+
 #
 # Decorators to validate state.
 #
@@ -937,9 +958,12 @@ class ObjectVersion(db.Model, Timestamp):
         actual data on disk is not copied. Instead, the two object versions
         will point to the same physical file (via the same FileInstance).
 
+        All the tags associated with the current object version are copied over
+        to the new instance.
+
         .. warning::
 
-           If the destination object exists, it will be replaced by  the new
+           If the destination object exists, it will be replaced by the new
            object version which will become the latest version.
 
         :param bucket: The bucket (instance or id) to copy the object to.
@@ -948,11 +972,16 @@ class ObjectVersion(db.Model, Timestamp):
             Default: current object key.
         :returns: The copied object version.
         """
-        return ObjectVersion.create(
+        o = ObjectVersion.create(
             self.bucket if bucket is None else as_bucket(bucket),
             key or self.key,
             _file_id=self.file_id
         )
+
+        for t in self.tags:
+            t.copy(object_version=o)
+
+        return o
 
     @ensure_unlocked(getter=lambda o: not o.bucket.locked)
     def remove(self):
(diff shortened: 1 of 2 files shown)


jirikuncar:
Why do we need another key, value store? I thought that there was an agreement on storing additional metadata in records. Are you using it for filtering? Wouldn't be sufficient to have just a JSON field?

invenio_files_rest/models.py
```diff
@@ -1111,6 +1121,112 @@ def relink_all(cls, old_file, new_file):
             ObjectVersion.query.filter_by(file_id=str(old_file.id)).update({
                 ObjectVersion.file_id: str(new_file.id)})
 
+    def get_tags(self):
```
egabancho: Because I copy/paste the behavior from [BucketTags](https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/models.py#L399) 😉

egabancho:
> Why do we need another key, value store? I thought that there was an agreement on storing additional metadata in records

The other key, value store we have is at the level of the bucket, sometimes one needs to store some technical metadata with the file, which doesn't belong in the record metadata, i.e. the bitrate of a video file. 
Indeed we thought that it would be enough to store it in the record but as we work more with it, we see that it is more convenient and seems more natural to have it at level of the `ObjectVersion` instead.

> Wouldn't be sufficient to have just a JSON field?

Do you mean adding a JSON field to the current ObjectVersion? We discuss this option, but I think it is better to keep the same structure (I like consistency) across the file storage module, Bucket - BucketTags, ObjectVersion - ObjectVersionTags

> It reminds be the `more_info` mapping from BibDocFiles.

Well, yes, but I think it reminds Invenio 1.X time in the same way as [invenio-indexer](https://github.com/inveniosoftware/invenio-indexer/blob/master/invenio_indexer/tasks.py#L35) reminds `BibIndex`.
Anyhow it seems that we are not the only ones thinking about that, [amazon S3](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-metadata) has something similar and also [Backblaze](https://www.backblaze.com/b2/docs/files.html) (although this last one is lest mature)

egabancho:
It looks like it needs https://github.com/inveniosoftware/invenio-accounts/pull/167 to pass the tests.

invenio_files_rest/models.py
```diff
@@ -1111,6 +1121,112 @@ def relink_all(cls, old_file, new_file):
...
+
+class ObjectVersionTag(db.Model):
+    """Model for storing tags associated to object versions.
+
+    This is useful to store extra technical information for an object version.
+    """
+
+    __tablename__ = 'files_objecttags'
+
+    object_version_bucket_id = db.Column(
```
lnielsen: Since these tags relate to a specific object version you can remove the bucket and key and just link them via ``ObjectVersion.version_id``. Naming-wise I would go with ``version_id`` instead of ``object_version_id`` (similar to naming used in MultipartObject).

lnielsen:
> Why do we need another key, value store?

I see a use case for ``ObjectTags`` for technical file metadata (files-rest) as opposed to descriptive metadata (record).

> Wouldn't be sufficient to have just a JSON field?

I wouldn't go with a JSON field to ensure that we can align with S3: 

- http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-metadata (user defined metadata)
- http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html (``x-amz-meta-*``)

invenio_files_rest/models.py
egabancho: This is what I did at the beginning, but SQLAlchemy doesn't seem to like it.
I'll change the naming though.

invenio_files_rest/models.py
egabancho: I am not sure how to use the `association_proxy` ...
Looking at the docs it seems to be used in many-to-many relationships, whereas here we have a one-to-many.
I tried setting it up by it makes things quite complex adding some nasty circular dependencies between `ObjectVersion` and `ObjectVersionTag`

egabancho:
As I discuss IRL with @lnielsen, I have changed the primary key of `ObjectVersion` to be only `version_id` and have added a unique constraint to replace the old primary key

egabancho:
ping 😉

jirikuncar:
FYI this is a good candidate that could benefit from an Alembic upgrade recipe.

egabancho:
🤔  I didn't know we are using Alembic now ...
Anyhow, should we make migration recipes for packages in *alpha* versions?
I thought the whole point of releasing packages in their early stages was to be able to be more agile in this aspect.

invenio_files_rest/models.py
```diff
@@ -946,11 +972,18 @@ def copy(self, bucket=None, key=None):
...
-            key or self.key,
-            _file_id=self.file_id
-        )
+        with db.session.begin_nested():
+            o = ObjectVersion.create(
+                self.bucket if bucket is None else as_bucket(bucket),
+                key or self.key,
+                _file_id=self.file_id
+            )
+            db.session.add(o)
```
lnielsen: ``ObjectVersion.create`` is already running in a nested transaction so it should not be necessary do it here as well (i.e. remove ``with db.session.begin_nested()`` + ``db.session.add(o)``.

jirikuncar:
@egabancho many packages are already running in production and we not released only due to lack of maintainers time. I see that the milestone is set to v1.0.0 so we can wait until all packages are in beta (I3B) with initial upgrade recipes.

tests/test_models.py
```diff
@@ -791,3 +791,49 @@ def test_fileinstance_validation(app, db, dummy_location):
     f = FileInstance.create()
     f.set_uri('x' * 255, 1000, 1000)  # Should not raise
     pytest.raises(ValueError, f.set_uri, 'x' * 256, 1000, 1000)
+
+
+def test_object_version_tags(app, db, dummy_location):
```
lnielsen: I can see you just copy/pasted the tests from buckets, but I think the person (i.e. me) who wrote the bucket tests was a bit lazy :-) Would you mind improving them a bit. E.g.:

- Use the ``objects`` test fixture.
- Split into multiple tests (if it makes sense - hopefully with the test fixtures this shouldn't add much extra code.

tests/test_models.py
```diff
@@ -791,3 +791,49 @@ def test_fileinstance_validation(app, db, dummy_location):
...
+
+    # Test get
+    assert ObjectVersionTag.query.count() == 2
+    assert ObjectVersionTag.get(obj1, "mykey").value == "testvalue"
+    assert ObjectVersionTag.get_value(obj1.version_id, "another_key") \
+        == "another value"
+    assert ObjectVersionTag.get_value(obj1, "invalid") is None
+
+    # Test delete
+    ObjectVersionTag.delete(obj1, "mykey")
```
lnielsen: Can you also add a test for the cascade of ``ObjectVersion.delete()``. I.e. that if you delete an ``ObjectVersion``, the tags are also deleted.

tests/test_models.py
```diff
@@ -791,3 +791,49 @@ def test_fileinstance_validation(app, db, dummy_location):
...
+    db.session.commit()
+    b = Bucket.create()
+    obj1 = ObjectVersion.create(b, "test").set_file(f)
+    ObjectVersionTag.create(obj1, "mykey", "testvalue")
+    ObjectVersionTag.create(obj1, "another_key", "another value")
+    db.session.commit()
+
+    # Duplicate key
+    pytest.raises(
+        Exception, ObjectVersionTag.create, obj1, "mykey", "newvalue")
```
lnielsen: Can the exception be made more specific - e.g. is it ``IntegrityError`` being thrown? Would you mind also fixing it in the BucketTags test.

tests/test_models.py
egabancho: AFAIK `ObjectVersion.delete()` does not delete the `ObjectVersion` literally from the db (from the docstring) ` Technically works by creating a new version which works as a delete marker.` 

I did `ObjectVersion.query.delete()` instead, which I guess is what you were thinking in the first place 😉

egabancho:
https://github.com/inveniosoftware/invenio-files-rest/pull/147

PR #138, opened 2016-11-24
views: bad response code fix

lnielsen:
* Fixes issue with GET request to LocationResource replying with 500
  error instead of 405.

Signed-off-by: Lars Holm Nielsen <lars.holm.nielsen@cern.ch>


jirikuncar:
LGTM

PR #147, opened 2017-04-11

egabancho:
* Introduces new model to store tags (key,value pairs) associated with
a given object version.


krzysztof:
Should this PR be closed since https://github.com/inveniosoftware/invenio-files-rest/pull/167 supercedes it?

PR #156, opened 2017-06-26
Header tags

nikofil:


invenio_files_rest/views.py
```diff
@@ -101,7 +104,7 @@ def invalid_subresource_validator(value):
         load_from='Content-MD5',
         location='headers',
     ),
-    'x_invenio_tags': fields.Str(
+    'x_invenio_tags': QueryString(
```
egabancho: I think it would make more sense if we call it something like `x-invenio-files-tagging`, just to add a valid namespace to the header key.

egabancho:
@nikofil I would very much appreciate if you could take a look at the tests 😃 , maybe @dinosk can help you with the alembic errors.

invenio_files_rest/views.py
```diff
@@ -523,6 +547,11 @@ def create_object(self, bucket, key):
             obj = ObjectVersion.create(bucket, key)
             obj.set_contents(
                 stream, size=content_length, size_limit=size_limit)
+            # Check add/update tags
+            if tags:
+                for key, value in tags.items():
+                    ObjectVersionTag.create_or_update(obj, key, value)
```
egabancho: This should be `create` instead.

invenio_files_rest/views.py
```diff
@@ -802,6 +831,47 @@ def delete(self, bucket=None, key=None, version_id=None, upload_id=None,
...
+
+class TagResource(ContentNegotiatedMethodView):
+    """Tag resource."""
+
+    def __init__(self, *args, **kwargs):
+        """Instantiate content negotiated view."""
+        super(TagResource, self).__init__(*args, **kwargs)
+
+    @pass_bucket
+    def put(self, key=None, bucket=None):
```
egabancho: I think we should add some sort of access rights check, i.e. `@need_bucket_permission('bucket-update')`.

tests/test_views_objectversion.py
```diff
@@ -502,3 +502,30 @@ def test_delete_unwritable(client, db, bucket, versions):
...
+def test_put_header_tags(client, bucket, permissions, get_md5, get_json):
+    """Test upload of an object with tags in the headers."""
+
+    key = 'test.txt'
+    data = b'updated_content'
+    checksum = get_md5(data, prefix=True)
+    object_url = url_for(
+        'invenio_files_rest.object_api', bucket_id=bucket.id, key=key)
+    headers = {
+        'X-Invenio-Tags': 'key1=val1;key2=val2;key3=val3;key1=invalid'
```
egabancho: This actually should fail because you are repeating one of the keys. I would also change the formatting to follow what amazon uses `tag1=value1&tag2=value2`

tests/test_views_objectversion.py
```diff
@@ -502,3 +502,30 @@ def test_delete_unwritable(client, db, bucket, versions):
...
+    resp = client.put(
+        object_url,
+        input_stream=BytesIO(data),
+        headers=headers,
+    )
+    assert resp.status_code == 200
+    assert resp.get_etag()[0] == checksum
+
+    tags = ObjectVersion.get(bucket, key).get_tags()
+    assert tags['key1'] == 'val1'
```
egabancho: It is funny, with you current implementation this test should fail 😉 .

dinosk:
@nikofil hey Niko, you just have to add the new table that is created in the models in an alembic revision. The documentation for it is here https://github.com/inveniosoftware/invenio-db/blob/master/alembic.rst check the `invenio alembic revision` part which does it automatically.

egabancho:
Superseded by https://github.com/inveniosoftware/invenio-files-rest/pull/167

PR #158, opened 2017-08-04
Checksum Verification Scheduler

slint:


invenio_files_rest/tasks.py
```diff
@@ -50,16 +55,109 @@ def progress_updater(size, total):
...
+    files = files.yield_per(1000)
+    scheduled_file_ids = []
+    total_size = 0
+    for f in files:
+        # Add at least the first file, since it might be larger than "max_size"
+        scheduled_file_ids.append(str(f.id))
+        total_size += f.size
+        if max_size and max_size < total_size:
+            break
+    print('Sending this files for checksum:', scheduled_file_ids)
```
lnielsen: Is the print supposed to be here? If yes, then perhaps better to use the task logger.

slint: Slipped during debug. Removed and pushed some tests to pass the coverage as well

lnielsen:
Test passes on lowest requirements, so I think it's an Celery v3 vs v4 issue.

slint:
@lnielsen Everything passing now. There's no easy way to test for warnings being logged though...

PR #161, opened 2017-08-09
test: use mysql 5.7 for travis tests

nharraud:
* Replaces mysql 5.6 with mysql 5.7 on travis because of a
  regression bug. (see https://bugs.mysql.com/bug.php?id=77591)

Signed-off-by: Nicolas Harraudeau <nicolas.harraudeau@cern.ch>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ cache:
     - $HOME/.cache/pip
 
 services:
-  - mysql
+  # - mysql
   - postgresql
   - redis
   - rabbitmq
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -47,11 +47,13 @@ dist: trusty
 sudo: required
 addons:
    postgresql: "9.4"
-   apt:
-    packages:
-    - mysql-server-5.6
-    - mysql-client-core-5.6
-    - mysql-client-5.6
+   mysql: "5.7"
```
krzysztof: Perhaps just specifying mysql 5.7 would be enough? (no need for the custom install script above) (@slint)

slint: In `invenio-acccess`, mysql also just openly specified: https://github.com/inveniosoftware/invenio-access/blob/master/.travis.yml#L42

@lnielsen At this point eg. Zenodo is running postgres 9.4, but shouldn't we pin specific versions of DBs/services just to have some guarantee of minimum requirements?

nharraud: @krzysztof I added it to test based on @hachreak code https://github.com/hachreak/invenio-db/blob/json-support/.travis.yml.
When you look at what version travis starts it is still 5.6. So it seems that this like has no effect.

nharraud: @slint Yes we should say what versions Invenio has been tested with. It is however hard to say it now as Invenio is still unstable and I am not sure that all services, i.e. CDS labs, Inspire, Zenodo & B2SHARE, use the same versions.

nharraud: I am currently searching for a way to install mysql without needing root access. This should make travis faster as we can then set `sudo: false`

slint: To be "precise", it's a combination of specifying `dist: precise` and plain `mysql` (no version) which by default installs `mysql v5.5`... https://github.com/inveniosoftware/invenio-access/blob/master/.travis.yml#L26

nharraud:
**Notes:**
The following solution doesn't work for some reason:
```
addons:
  apt:
     sources:
       - mysql-5.7-trusty This doesn't work for some reason
     packages:
         - mysql-community-client
         - mysql-community-server
```
see travis-ci/travis-ci#7992

nharraud:
Closing this issue. We would have to use mysql-5-7 in every module which is not acceptable given how travis is slow. We will instead change the [alembic recipe in invenio-access](https://github.com/inveniosoftware/invenio-access/blob/fc32cf2a522f0c464715f9998d1a6cfd11a3eef6/invenio_access/alembic/dbe499bfda43_prevent_null_in_actionusers_for_.py#L38) so that it does not trigger [this mysql bug](https://github.com/inveniosoftware/invenio-access/blob/fc32cf2a522f0c464715f9998d1a6cfd11a3eef6/invenio_access/alembic/dbe499bfda43_prevent_null_in_actionusers_for_.py#L38).

PR #167, opened 2017-08-29

hachreak:
* Changes primary constrain from multiple columns to only
  ``version_id``.

* Introduces new model to store tags (key,value pairs) associated with
  a given object version. (closes #147)

### DIFF
--- a/invenio_files_rest/alembic/8ae99b034410_create_files_objecttags_table.py
+++ b/invenio_files_rest/alembic/8ae99b034410_create_files_objecttags_table.py
@@ -21,7 +21,7 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
-"""Revision message."""
+"""Create files_objecttags table."""
 
 import sqlalchemy as sa
 import sqlalchemy_utils
--- a/invenio_files_rest/alembic/8ae99b034410_revision_message.py
+++ b/invenio_files_rest/alembic/8ae99b034410_revision_message.py
@@ -1,0 +1,99 @@
+#
+# This file is part of Invenio.
+# Copyright (C) 2017 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Revision message."""
+
+import sqlalchemy as sa
+import sqlalchemy_utils
+from alembic import op
+from sqlalchemy.dialects import mysql
+
+# revision identifiers, used by Alembic.
+revision = '8ae99b034410'
+down_revision = 'f741aa746a7d'
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    # table ObjectVersion: modify primary_key
+    if op.get_context().dialect.name == 'mysql':
+        Fk = 'fk_files_object_bucket_id_files_bucket'
+        op.execute(
+            'ALTER TABLE files_object '
+            'DROP FOREIGN KEY {0}, DROP PRIMARY KEY, '
+            'ADD PRIMARY KEY(version_id), '
+            'ADD FOREIGN KEY(bucket_id) '
+            'REFERENCES files_bucket(id)'.format(Fk))
+    else:
+        op.drop_constraint('pk_files_object', 'files_object', type_='primary')
+        op.create_primary_key(
+            'pk_files_object', 'files_object', ['version_id'])
+    op.create_unique_constraint(
+        'uq_files_object_bucket_id', 'files_object',
+        ['bucket_id', 'version_id', 'key'])
+    # table ObjectVersionTag: create
+    op.create_table(
+        'files_objecttags',
+        sa.Column(
+            'version_id',
+            sqlalchemy_utils.types.uuid.UUIDType(),
+            nullable=False),
+        sa.Column(
+            'key',
+            sa.Text().with_variant(mysql.VARCHAR(255), 'mysql'),
+            nullable=False
+        ),
+        sa.Column(
+            'value',
+            sa.Text().with_variant(mysql.VARCHAR(255), 'mysql'),
+            nullable=False
+        ),
+        sa.PrimaryKeyConstraint('version_id', 'key'),
+        sa.ForeignKeyConstraint(
+            ['version_id'],
+            [u'files_object.version_id'],
+            ondelete='CASCADE'),
+    )
+
+
+def downgrade():
+    """Downgrade database."""
+    # table ObjectVersionTag
+    op.drop_table('files_objecttags')
+    # table ObjectVersion: modify primary_key
+    if op.get_context().dialect.name == 'mysql':
+        Fk = 'fk_files_object_bucket_id_files_bucket'
+        op.execute(
+            'ALTER TABLE files_object '
+            'DROP INDEX uq_files_object_bucket_id, '
+            #  'DROP FOREIGN KEY {0}, '
+            'DROP PRIMARY KEY, '
+            'ADD PRIMARY KEY(`bucket_id`, `key`, `version_id`)'
+            .format(Fk))
+    else:
+        op.drop_constraint(
+            'pk_files_object', 'files_object', type_='primary')
+        op.create_primary_key('pk_files_object', 'files_object',
+                              ['bucket_id', 'key', 'version_id'])
--- a/invenio_files_rest/models.py
+++ b/invenio_files_rest/models.py
@@ -96,7 +96,7 @@ def validate_key(key):
 
 
 def as_bucket(value):
-    """Get a bucket object from a bucket id or bucket.
+    """Get a bucket object from a bucket ID or a bucket object.
 
     :param value: A :class:`invenio_files_rest.models.Bucket` or a Bucket ID.
     :returns: A :class:`invenio_files_rest.models.Bucket` instance.
@@ -105,7 +105,7 @@ def as_bucket(value):
 
 
 def as_bucket_id(value):
-    """Get a bucket id from a bucket id or bucket.
+    """Get a bucket ID from a bucket ID or a bucket object.
 
     :param value: A :class:`invenio_files_rest.models.Bucket` instance of a
         bucket ID.
@@ -115,7 +115,7 @@ def as_bucket_id(value):
 
 
 def as_object_version(value):
-    """Get a object version object from a object version id or object version.
+    """Get an object version object from an object version ID or an object version.
 
     :param value: A :class:`invenio_files_rest.models.ObjectVersion` or an
         object version ID.
@@ -126,10 +126,10 @@ def as_object_version(value):
 
 
 def as_object_version_id(value):
-    """Get a object version id from a object version id or object version.
+    """Get an object version ID from an object version ID or an object version.
 
     :param value: A :class:`invenio_files_rest.models.ObjectVersion` instance
-        of a object version id.
+        of a object version ID.
     :returns: The :class:`invenio_files_rest.models.ObjectVersion` version_id.
     """
     return value.version_id if isinstance(value, ObjectVersion) else value
@@ -450,7 +450,7 @@ class Bucket(db.Model, Timestamp):
         for o in ObjectVersion.get_heads_by_bucket(self):
             o.merge(bucket=bucket, key=o.key)
 
-        bucket.locked = True if lock else self.locked
+        bucket.locked = lock or self.locked
 
         return bucket
 
@@ -462,9 +462,9 @@ class Bucket(db.Model, Timestamp):
     def create(cls, location=None, storage_class=None, **kwargs):
         r"""Create a bucket.
 
-        :param location: Location of bucket (instance or name).
+        :param location: Location of a bucket (instance or name).
             Default: Default location.
-        :param storage_class: Storage class of bucket.
+        :param storage_class: Storage class of a bucket.
             Default: Default storage class.
         :param \**kwargs: Keyword arguments are forwarded to the class
             constructor.
@@ -487,7 +487,7 @@ class Bucket(db.Model, Timestamp):
 
     @classmethod
     def get(cls, bucket_id):
-        """Get bucket object (excluding deleted).
+        """Get a bucket object (excluding deleted).
 
         :param bucket_id: Bucket identifier.
         :returns: Bucket instance.
@@ -1298,7 +1298,7 @@ class ObjectVersion(db.Model, Timestamp):
 class ObjectVersionTag(db.Model):
     """Model for storing tags associated to object versions.
 
-    This is useful to store extra technical information for an object version.
+    Used for storing extra technical information for an object version.
     """
 
     __tablename__ = 'files_objecttags'
@@ -1336,7 +1336,7 @@ class ObjectVersionTag(db.Model):
 
     @classmethod
     def get(cls, object_version, key):
-        """Get tag object."""
+        """Get the tag object."""
         return cls.query.filter_by(
             version_id=as_object_version_id(object_version),
             key=key,
@@ -1354,7 +1354,7 @@ class ObjectVersionTag(db.Model):
 
     @classmethod
     def create_or_update(cls, object_version, key, value):
-        """Create or update a new tag for given object version."""
+        """Create or update a new tag for a given object version."""
         obj = cls.get(object_version, key)
         if obj:
             obj.value = value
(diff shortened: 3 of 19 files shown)


hachreak:
ping @lnielsen @egabancho :shipit: ?

invenio_files_rest/views.py
```diff
@@ -802,6 +835,49 @@ def delete(self, bucket=None, key=None, version_id=None, upload_id=None,
...
+
+class TagResource(ContentNegotiatedMethodView):
+    """Tag resource."""
+
+    def __init__(self, *args, **kwargs):
+        """Instantiate content negotiated view."""
+        super(TagResource, self).__init__(*args, **kwargs)
+
+    @pass_bucket
+    @need_bucket_permission('bucket-update')
```
krzysztof: @lnielsen suggested that tags need a finer permission granularity (``tag-update``)

invenio_files_rest/models.py
```diff
@@ -422,7 +422,7 @@ def merge(self, bucket, lock=False):
         for o in ObjectVersion.get_heads_by_bucket(self):
             o.merge(bucket=bucket, key=o.key)
 
-        bucket.locked = True if lock else self.locked
+        bucket.locked = lock or self.locked
```
krzysztof: This is the only non-docs change.

invenio_files_rest/alembic/8ae99b034410_revision_message.py
```diff
@@ -0,0 +1,99 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Revision message."""
```
krzysztof: Would be good to name it better (including this filename).

invenio_files_rest/views.py
hachreak: do you think it's better something like `objectversion-tag-update`?

krzysztof: @hachreak yup

krzysztof: especially w.r.t. https://github.com/inveniosoftware/invenio-files-rest/pull/167#discussion_r138885083

krzysztof:
@egabancho Will you have time to work on this?

ntarocco:
We are gonna work on this branch to merge it, I will add a comment.

invenio_files_rest/models.py
```diff
@@ -1156,6 +1233,118 @@ def relink_all(cls, old_file, new_file):
...
+        return not self.__eq__(other=other)
+
+    def merge(self, bucket, key=None):
+        """Merge the object inside the bucket."""
+        # check no new version is added while the object is modified
+        orig_ancestor = ObjectVersion.get_versions(
+            bucket=self.bucket, key=key, desc=True)
+        dest_head = ObjectVersion.query_heads_by_bucket(bucket=bucket).filter(
+            ObjectVersion.key == key).one_or_none()
+        if dest_head and all(orig.file_id != dest_head.file_id
```
ntarocco: Comment for us when we re-check the code:
This does not ensure that a new version is not added immediately after by another thread or process.
Is it "safe enough" like this, or do we want to introduce an "atomic transaction"?

invenio_files_rest/views.py
```diff
@@ -101,20 +104,27 @@ def invalid_subresource_validator(value):
         load_from='Content-MD5',
         location='headers',
     ),
+    'x_invenio_file_tags': QueryString(
+        load_from='X-Invenio-File-Tags',
```
jirikuncar: I would remove the `X-` prefix.

switowski:
I think I applied all of the suggestions from the comments.
Is there anything else stopping this commit from being merged?
ping @lnielsen @egabancho @krzysztof

invenio_files_rest/models.py
```diff
@@ -408,27 +429,30 @@ def validate_storage_class(self, key, default_storage_class):
...
-        for o in ObjectVersion.get_by_bucket(self):
-            o.copy(bucket=b)
+    @ensure_not_deleted()
+    def merge(self, bucket, lock=False):
+        """Sync self to the destination bucket."""
+        for o in ObjectVersion.get_heads_by_bucket(self):
+            o.merge(bucket=bucket, key=o.key)
 
-        b.locked = True if lock else self.locked
+        bucket.locked = lock or self.locked
```
egabancho: I don't know which should be the correct way of using the `lock` parameter, if I set it to `False` I think that the returning bucket should be `bucket.locked = False`.
Otherwise you should always do
```python
new_bucket = bucket.snapshot()
new_bucket.locked = False
```
@switowski WDYT?

invenio_files_rest/models.py
```diff
@@ -1187,6 +1264,118 @@ def relink_all(cls, old_file, new_file):
...
+        return obj
+
+    @classmethod
+    def create_or_update(cls, object_version, key, value):
+        """Create or update a new tag for a given object version."""
+        obj = cls.get(object_version, key)
+        if obj:
+            obj.value = value
+        else:
+            cls.create(object_version, key, value)
```
egabancho: we should return the new tag as create does
```python
obj = cls.get(object_version, key
if obj:
    obj.value = value
    db.session.merge(obj)
else:
    obj = cls.create(object_version, key, value)
return obj
```

invenio_files_rest/models.py
egabancho: You are right, probably a `with db.session.begin_nested():` wouldn't hurt that much.

invenio_files_rest/models.py
```diff
@@ -114,6 +114,27 @@ def as_bucket_id(value):
     return value.id if isinstance(value, Bucket) else value
 
 
+def as_object_version(value):
```
lnielsen: This method is unused as far as I can see.

invenio_files_rest/models.py
```diff
@@ -1020,11 +1049,35 @@ def copy(self, bucket=None, key=None):
             Default: current object key.
         :returns: The copied object version.
         """
-        return ObjectVersion.create(
-            self.bucket if bucket is None else as_bucket(bucket),
-            key or self.key,
-            _file_id=self.file_id
-        )
+        # get (if exists) the latest version
```
lnielsen: I think the ``copy()`` method will need some refactoring. As it is now, it will execute quite a lot of extra queries compared to the previous version. It seems the extra check to not copy an object if it already exists, was done because of the merge method. If so, perhaps the complexity can be moved to the merge method. Alternatively, we could add e.g. two parameters ``copy(self ,..., force=True, with_tags=False)`` that would control if 1) an object is copied if it already exists and 2) if tags are copied.

invenio_files_rest/models.py
```diff
@@ -1157,19 +1212,41 @@ def delete(cls, bucket, key):
...
+
+    @classmethod
+    def query_heads_by_bucket(cls, bucket):
+        """Create a query to get only HEAD filtered by bucket."""
+        return cls.query_by_bucket(bucket=bucket).filter(
+            cls.is_head.is_(True)
+        )
+
+    @classmethod
+    def get_heads_by_bucket(cls, bucket):
```
lnielsen: ditto here

invenio_files_rest/models.py
```diff
@@ -1187,6 +1264,118 @@ def relink_all(cls, old_file, new_file):
...
+    def __eq__(self, other):
+        """Check if the two object are equals."""
+        return other and isinstance(other, self.__class__) and \
+            self.key == other.key and self.file_id == other.file_id
+
+    def __ne__(self, other):
+        """Check if are not equal."""
+        return not self.__eq__(other=other)
+
+    def merge(self, bucket, key=None):
```
lnielsen: I'm not sure I fully understand the use case for the merging support. It seems to add quite a lot of complexity overall. To discuss IRL.

invenio_files_rest/models.py
```diff
@@ -1187,6 +1264,118 @@ def relink_all(cls, old_file, new_file):
...
+            cls.create(object_version, key, value)
+
+    @classmethod
+    def get_value(cls, object_version, key):
+        """Get the tag value."""
+        obj = cls.get(object_version, key)
+        return obj.value if obj else None
+
+    @classmethod
+    def delete(cls, object_version, key):
```
lnielsen: I would add support to delete all keys as well. Essentially allow either not specifying ``key`` as a kwarg or pass None.

invenio_files_rest/utils.py
```diff
@@ -23,9 +23,14 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
 """Implementention of various utility functions."""
+try:
```
lnielsen: Use ``six`` instead as it's already a dependency.

invenio_files_rest/views.py
```diff
@@ -529,6 +549,11 @@ def create_object(self, bucket, key):
             obj = ObjectVersion.create(bucket, key)
             obj.set_contents(
                 stream, size=content_length, size_limit=size_limit)
+            # Check add tags
```
lnielsen: It does not seems there's any checking of permission to create tags.

invenio_files_rest/views.py
```diff
@@ -814,6 +839,49 @@ def delete(self, bucket=None, key=None, version_id=None, upload_id=None,
...
+
+class ObjectVersionTagResource(ContentNegotiatedMethodView):
+    """Tag resource."""
+
+    def __init__(self, *args, **kwargs):
+        """Instantiate content negotiated view."""
+        super(ObjectVersionTagResource, self).__init__(*args, **kwargs)
+
+    @pass_bucket
+    @need_bucket_permission('bucket-update')
```
lnielsen: Special actions like ``files-rest-objectversion-tag-update`` have been added in ``permissions.py`` but are not used here

invenio_files_rest/views.py
```diff
@@ -814,6 +839,49 @@ def delete(self, bucket=None, key=None, version_id=None, upload_id=None,
...
+            db.session.commit()
+        return self.make_response(
+            data=obj,
+            context={
+                'class': ObjectVersion,
+            }
+        )
+
+    @pass_bucket
+    @need_bucket_permission('bucket-update')
```
lnielsen: Ditto

invenio_files_rest/views.py
```diff
@@ -814,6 +839,49 @@ def delete(self, bucket=None, key=None, version_id=None, upload_id=None,
             obj = self.get_object(bucket, key, version_id)
             return self.delete_object(bucket, obj, version_id)
 
+
+class ObjectVersionTagResource(ContentNegotiatedMethodView):
```
lnielsen: Add support for ``get()`` as well.

invenio_files_rest/views.py
```diff
@@ -814,6 +839,49 @@ def delete(self, bucket=None, key=None, version_id=None, upload_id=None,
...
+    def put(self, key=None, bucket=None):
+        """Create or update tags."""
+        tags = request.get_json()
+        obj = ObjectVersion.get(bucket, key)
+        if tags is not None:
+            with db.session.begin_nested():
+                for key, value in tags.items():
+                    ObjectVersionTag.create_or_update(obj, key, value)
+            db.session.commit()
+        return self.make_response(
```
lnielsen: Seems like a dedicated serializer/deserializer is missing for tags.

setup.py
```diff
@@ -33,14 +33,14 @@
...
-    'Flask-Menu>=0.4.0',
+    'Flask-Menu>=0.5.0',
     'SQLAlchemy-Continuum>=1.2.1',
     'check-manifest>=0.25',
     'coverage>=4.0',
     'invenio-access>=1.0.0a11',
-    'invenio-accounts>=1.0.0b3',
-    'invenio-admin>=1.0.0b1',
-    'invenio-celery>=1.0.0b2',
+    'invenio-accounts>=1.0.0b9',
```
lnielsen: Bump to latest released versions.

invenio_files_rest/alembic/8ae99b034410_create_files_objecttags_table.py
```diff
@@ -0,0 +1,99 @@
...
+from sqlalchemy.dialects import mysql
+
+# revision identifiers, used by Alembic.
+revision = '8ae99b034410'
+down_revision = 'f741aa746a7d'
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
```
lnielsen: Note, I've not tested these.

invenio_files_rest/models.py
```diff
@@ -1020,11 +1049,35 @@ def copy(self, bucket=None, key=None):
...
+
+        return latest_obj
+
+    def synchronize_tags(self, object_version):
+        """Sync tags of the two objects."""
+        # create/update all source tags
+        for t in self.tags:
+            ObjectVersionTag.create_or_update(object_version=object_version,
+                                              key=t.key, value=t.value)
+        # remove tags that are in destionation but not in source
```
lnielsen: ``destionation`` -> ``the destination``

lnielsen:
I've moved the PR to #183 to have a branch in the repository instead of Leo's private branch.

invenio_files_rest/models.py
ntarocco: We are heavily using it in videos: https://github.com/CERNDocumentServer/cds-videos/blob/cdslabs_qa/cds/modules/deposit/api.py#L47

invenio_files_rest/models.py
ntarocco: Agree, the behaviour has been changed here: before, a copy was created in any case in `create()` method (if the same object version was already there, a new identical version was added as head), while now only if not already present.
Not sure an extra param is what we need. I have the feeling some sort of optimization code was added... Shall we decide what is best?

invenio_files_rest/models.py
ntarocco: In videos, the merge method is called when re-publishing an already published deposit after edition. I guess this happens for a simple use case: we have transcoding videos files happening even if the record is published because it is a long process and we want to allow the generated video files to be added to the bucket after publishing.
So when you edit, we want to take the deposit bucket and merge it with the published one that might have new files.

Unfortunately, it is also causing issues. To be discussed yes.

invenio_files_rest/views.py
```diff
@@ -110,11 +113,15 @@ def default_partfactory(part_number=None, content_length=None,
     :param content_length: The content length. (Default: ``None``)
     :param content_type: The HTTP Content-Type. (Default: ``None``)
     :param content_md5: The content MD5. (Default: ``None``)
+    :param file_tags_header: The file tags. (Default: ``None``)
     :returns: The content length, the part number, the stream, the content
         type, MD5 of the content.
     """
+    header = current_app.config['FILES_REST_FILE_TAGS_HEADER']
```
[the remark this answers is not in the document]
ntarocco: I agree, and it looks to me where are not using at all the parts upload, but the upload `/api/files/<bucket>/key`, so for me we can remove it.

invenio_files_rest/views.py
ntarocco: Can we discuss this? I am not sure what you have in mind.

invenio_files_rest/views.py
ntarocco: Looks like a wrong copy paste to me, I will fix permissions.

invenio_files_rest/views.py
ntarocco: True, I can add this.

invenio_files_rest/views.py
```diff
@@ -848,3 +922,7 @@ def delete(self, bucket=None, key=None, version_id=None, upload_id=None,
     '/<string:bucket_id>/<path:key>',
     view_func=object_view,
 )
+blueprint.add_url_rule(
+    '/<string:bucket_id>/<path:key>/tags',
```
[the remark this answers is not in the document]
ntarocco: I checked and there is no usage for this on videos at least. About the refactoring I am not sure what you mean exactly, so let's have a quick chat.

setup.py
ntarocco: I can do that, but then I might have a lot more to refactor in videos, to depend on new version. Quick discussion here too.

invenio_files_rest/models.py
```diff
@@ -1124,7 +1177,7 @@ def get(cls, bucket, key, version_id=None):
         return cls.query.filter(*filters).one_or_none()
 
     @classmethod
-    def get_versions(cls, bucket, key):
+    def get_versions(cls, bucket, key, desc=True):
```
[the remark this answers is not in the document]
ntarocco: Actually, looking at the code, I would not do that. Since this is returning a query, the code that is using this method can filter/sort/whatever. I am removing it and fixing its usage.

lnielsen:
I'm closing this PR in favor of #183 which moves the branch from private repo to this repo.

invenio_files_rest/models.py
```diff
@@ -408,27 +429,30 @@ def validate_storage_class(self, key, default_storage_class):
...
-            db.session.add(b)
+            db.session.add(bucket)
+        return self.merge(lock=lock, bucket=bucket)
 
-        for o in ObjectVersion.get_by_bucket(self):
-            o.copy(bucket=b)
+    @ensure_not_deleted()
+    def merge(self, bucket, lock=False):
+        """Sync self to the destination bucket."""
+        for o in ObjectVersion.get_heads_by_bucket(self):
```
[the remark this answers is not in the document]
ntarocco: @lnielsen `get_by_bucket(cls, bucket, versions=False)` is already returning heads when versions is False, we can just use that...

invenio_files_rest/alembic/8ae99b034410_create_files_objecttags_table.py
```diff
@@ -0,0 +1,99 @@
...
+    """Upgrade database."""
+    # table ObjectVersion: modify primary_key
+    if op.get_context().dialect.name == 'mysql':
```
ntarocco: For info, without the MySQL specific code, there is this error:
```
sqlalchemy.exc.InternalError: (pymysql.err.InternalError) (1025, "Error on rename of './mydb/#sql-1_4' to './mydb/files_object' (errno: 150 - Foreign key constraint is incorrectly formed)") [SQL: 'ALTER TABLE files_object DROP PRIMARY KEY '] (Background on this error at: http://sqlalche.me/e/2j85)
```

PR #169, opened 2017-09-27
models: no location timestamp update on bucket creation

jmartinm:
* Avoids updating location timestamps when creating buckets. This update
  is problematic when many concurrent tasks are creating buckets since
  the same row gets updated and can lead to deadlocks in the database.

Signed-off-by: Javier Martin Montull <javier.martin.montull@cern.ch>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,6 @@
 Changes
 =======
 
-Version 1.0.0a19 (released 2017-08-18)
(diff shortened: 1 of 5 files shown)


kaplun:
Travis is green, I guess if you can make a release commit then maybe @egabancho or others can make a release?

jmartinm:
Added release commit.

Issue #171, opened 2017-12-07
models: ensure_state creates errors that are not JSON serializable


jacquerie:
If we pass a string argument to a `RESTException` subclass, the error instance that we get is not JSON serializable:
```python
>>> from invenio_rest_files.errors import InvalidOperationError
>>> error = InvalidOperationError("Can't do something")
>>> error.get_body()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jacquerie/.envs/invenio-files-rest/local/lib/python2.7/site-packages/invenio_rest/errors.py", line 90, in get_body
    errors = self.get_errors()
  File "/home/jacquerie/.envs/invenio-files-rest/local/lib/python2.7/site-packages/invenio_rest/errors.py", line 77, in get_errors
    return [e.to_dict() for e in self.errors] if self.errors else None
AttributeError: 'str' object has no attribute 'to_dict'
```
But this is exactly what [`ensure_state`](https://github.com/inveniosoftware/invenio-files-rest/blob/983d0824a2ad199d5578d5df9b044bb6a7996e24/invenio_files_rest/models.py#L130-L141) is doing [here](https://github.com/inveniosoftware/invenio-files-rest/blob/983d0824a2ad199d5578d5df9b044bb6a7996e24/invenio_files_rest/models.py#L164-L167), [here](https://github.com/inveniosoftware/invenio-files-rest/blob/983d0824a2ad199d5578d5df9b044bb6a7996e24/invenio_files_rest/models.py#L179-L182) or [here](https://github.com/inveniosoftware/invenio-files-rest/blob/eb1aaa1b1c384e64110d0052d0a994030da12944/invenio_files_rest/models.py#L973), resulting in the exception described in https://github.com/inspirehep/inspire-next/issues/2983#issuecomment-349692877.

PR #172, opened 2017-12-07
models: JSON-friendly errors in ensure_state

jacquerie:
Closes https://github.com/inveniosoftware/invenio-files-rest/issues/171, but I'm not fully convinced that this is the best solution for the problem as all those small classes looks a bit clunky. It's also missing tests as they seem tricky to write, and if this solution is not going in the right direction then it would have been a waste of effort...

### DIFF
--- a/invenio_files_rest/models.py
+++ b/invenio_files_rest/models.py
@@ -141,6 +141,32 @@ def ensure_state(default_getter, exc_class, default_msg=None):
     return decorator
 
 
+class BucketError(object):
+    """Represents a bucket level error.
+
+    .. note:: This is not an actual exception.
+    """
+
+    def __init__(self, message):
+        self.res = dict(message=message)
+
+    def to_dict(self):
+        return self.res
+
+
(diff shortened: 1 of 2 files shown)


jacquerie:
> It's also missing tests as they seem tricky to write

Turns out they weren't *that* hard, so I added a PoC for them.

invenio_files_rest/models.py
```diff
@@ -141,6 +141,32 @@ def inner(self, *args, **kwargs):
     return decorator
 
 
+class BucketError(object):
```
lnielsen: Should they not inherit from ``Exception``?

jacquerie: These two classes are patterned over https://github.com/inveniosoftware/invenio-rest/blob/d5a3d185c0a8c8cd46860bd0fd9be8bb3786f157/invenio_rest/errors.py#L35-L58, which also doesn't inherit from `Exception`, because it's not intended to be thrown, but is just representing an error.

PR #189, opened 2019-02-12
models: adding unicode support

### LINKED ISSUE #178: __repr__ for ObjectVersion doesn't handle unicode
Probably changing [`ObjectVersion.__repr__`](https://github.com/inveniosoftware/invenio-files-rest/blob/74846ed4468365c72db6e18c34c9818c3ace5c34/invenio_files_rest/models.py#L899-L902) to smoething like:

```python
return u'{}:{}:{}'.format(...)
```

roksys:
* Adding unicode support for ObjectVersion. Closes #178

### DIFF
--- a/invenio_files_rest/models.py
+++ b/invenio_files_rest/models.py
@@ -38,6 +38,7 @@ from __future__ import absolute_import, print_function
 
 import mimetypes
 import re
+import sys
 import uuid
 from datetime import datetime
 from functools import wraps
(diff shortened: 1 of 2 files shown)


lnielsen:
Please have a look at the failing tests.

invenio_files_rest/models.py
```diff
@@ -964,11 +966,21 @@ def validate_key(self, key, key_):
         """Validate key."""
         return validate_key(key_)
 
-    def __repr__(self):
-        """Return representation of location."""
-        return '{0}:{2}:{1}'.format(
+    def __unicode__(self):
+        """Return unicoded object."""
+        return u"{0}:{2}:{1}".format(
```
okraskaj: is this order intentional?

invenio_files_rest/models.py
```diff
@@ -54,13 +54,15 @@
...
 import re
+import sys
 import uuid
 from datetime import datetime
 from functools import wraps
 from os.path import basename
 
 import six
 from flask import current_app
+from future.utils import python_2_unicode_compatible
```
Glignos: Leftover?

invenio_files_rest/models.py
```diff
@@ -964,10 +966,20 @@ def validate_key(self, key, key_):
         """Validate key."""
         return validate_key(key_)
 
-    def __repr__(self):
-        """Return representation of location."""
-        return '{0}:{2}:{1}'.format(
-            self.bucket_id, self.key, self.version_id)
+    def __unicode__(self):
```
Glignos: Instead you could use the `@python_2_unicode_compatible` decorator and just create a `__str__`  function returning unicode and the `__repr__` returning the `__str__` to avoid manually controlling the python version.

roksys: I wanted to use this decorator, but @slint suggested not to, since it is a class decorator, not a function one and could affect some other class behaviour.

diegodelemos:
@roksys this needs rebase :)

PR #202, opened 2019-02-21
mimetypes: improve detection

frankois:
Adds the following mimetypes:
- 'application/gzip'
- 'application/gzip'
- 'application/x-bzip2'
- 'application/x-xz'

### DIFF
--- a/invenio_files_rest/models.py
+++ b/invenio_files_rest/models.py
@@ -58,6 +58,7 @@ from .errors import BucketLockedError, FileInstanceAlreadySetError, \
     MultipartInvalidChunkSize, MultipartInvalidPartNumber, \
     MultipartInvalidSize, MultipartMissingParts, MultipartNotCompleted
 from .proxies import current_files_rest
+from .utils import ENCODING_MIMETYPES, guess_mimetype
 
 slug_pattern = re.compile('^[a-z][a-z0-9-]+$')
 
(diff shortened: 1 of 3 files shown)


invenio_files_rest/utils.py
```diff
@@ -35,3 +35,12 @@ def load_or_import_from_config(key, app=None, default=None):
     app = app or current_app
     imp = app.config.get(key)
     return obj_or_import_string(imp, default=default)
+
+
+def get_extra_encodings(encoding):
+    """Diego said to put documentation.
```
lnielsen: If Diego said so :D

invenio_files_rest/config.py
```diff
@@ -119,3 +119,11 @@
 
 FILES_REST_FILE_TAGS_HEADER = 'X-Invenio-File-Tags'
 """Header for updating file tags."""
+
+FILES_REST_EXTRA_ENCODINGS = {
```
lnielsen: I would not add this as a config, but simply hard code it in top of ``utils.py``

```python
ENCODING_MIMETYPES = {
    # ...
}
"""Mapping of encoding to MIME type."""

File: invenio_files_rest/utils.py [hunk 3]
```diff
@@ -35,3 +35,12 @@ def load_or_import_from_config(key, app=None, default=None):
     app = app or current_app
     imp = app.config.get(key)
     return obj_or_import_string(imp, default=default)
+
+
+def get_extra_encodings(encoding):
```
lnielsen: I would extra the entire function instead of only part of it.

```python
def guess_mimetype(filename):
    m, encoding = mimetypes.guess_type(filename)
    if encoding:
        m = ENCODING_MIMETYPES.get(encoding, None)
    return m or 'application/octet-stream'
```

invenio_files_rest/models.py
```diff
@@ -959,7 +960,10 @@ def mimetype(self):
         if self._mimetype:
             m = self._mimetype
         elif self.key:
-            m = mimetypes.guess_type(self.key)[0]
+            m, encoding = mimetypes.guess_type(self.key)
```
lnielsen: See comment below

```python
return self._mimetype if self._mimetype else guess_mimetype(self.key)
```

frankois:
Closed by : https://github.com/inveniosoftware/invenio-files-rest/pull/202

PR #207, opened 2019-05-15
views: detect if input stream was already read

### LINKED ISSUE #197: views: test upload of ``.json`` file via browser and invenio-files-js
We have reports from some users about Invenio-Files-REST does not allow upload of json files. This seems to be perhaps related to the browser somehow overriding the content-type.

diegodelemos:
* In the case of having an external library which has already
  read the `request.stream` files REST fails with a server error
  since it will not be able to read the uploaded file.

  This has happened with Sentry's `raven-python` library which
  reads the stream when the mimetype is `application/json` causing
  JSON files uploads to not work and throwing a misleading error.
  More here
  https://github.com/getsentry/raven-python/blob/719be9afae528962e683d086f9c2ceeda07f881b/raven/contrib/flask.py#L198

  Now a proper error is thrown (closes #197).

### DIFF
--- a/invenio_files_rest/views.py
+++ b/invenio_files_rest/views.py
@@ -267,6 +267,16 @@ def pass_multipart(with_completed=False):
     return decorate
 
 
+def ensure_input_stream_is_not_exhausted(f):
+    """Make sure that the input stream has not been read already."""
+    @wraps(f)
+    def decorate(*args, **kwargs):
+        if request.content_length and request.stream.is_exhausted:
+            abort(500)
+        return f(*args, **kwargs)
+    return decorate
+
+
 #
 # Permission checking
 #
@@ -797,6 +807,7 @@ class ObjectResource(ContentNegotiatedMethodView):
     @use_kwargs(post_args)
     @pass_bucket
     @need_bucket_permission('bucket-update')
+    @ensure_input_stream_is_not_exhausted
     def post(self, bucket=None, key=None, uploads=missing, upload_id=None):
         """Upload a new object or start/complete a multipart upload.
 
@@ -815,6 +826,7 @@ class ObjectResource(ContentNegotiatedMethodView):
     @use_kwargs(put_args)
     @pass_bucket
     @need_bucket_permission('bucket-update')
+    @ensure_input_stream_is_not_exhausted
     def put(self, bucket=None, key=None, upload_id=None):
         """Update a new object or upload a part of a multipart upload.
 
--- a/tests/test_views_multipart.py
+++ b/tests/test_views_multipart.py
@@ -543,3 +543,33 @@ def test_get_listuploads(client, db, bucket, multipart, multipart_url,
             bucket_id=str(bucket.id),
         ) + '?uploads')
         assert res.status_code == expected
+
+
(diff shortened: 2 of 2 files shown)


invenio_files_rest/views.py
```diff
@@ -267,6 +267,16 @@ def inner(self, bucket, key, upload_id, *args, **kwargs):
     return decorate
 
 
+def ensure_input_stream_is_not_read(f):
+    """Make sure that the input stream has not been read already."""
+    @wraps(f)
+    def decorate(*args, **kwargs):
+        if request.content_length and request.stream.is_exhausted:
+            abort(500)
```
ntarocco: Minor and IMHO: should we maybe add some more clear message? Please improve English! :)
```
abort(500, description="The request stream has been already consumed and cannot be read any more.")
```

invenio_files_rest/views.py
```diff
@@ -267,6 +267,16 @@ def inner(self, bucket, key, upload_id, *args, **kwargs):
     return decorate
 
 
+def ensure_input_stream_is_not_read(f):
```
ntarocco: Minor and IMHO: `ensure_input_stream_is_not_read` -> `ensure_input_stream_is_not_exhausted`?

Same for the test if changed.

tests/test_views_multipart.py
```diff
@@ -543,3 +543,33 @@ def test_get_listuploads(client, db, bucket, multipart, multipart_url,
...
+
+
+def test_already_read_input_stream(app, client, db, bucket, admin_user):
+    """Test server error when file stream is already read."""
+    key = 'test.json'
+    data = b'{"json": "file"}'
+    object_url = url_for(
+        'invenio_files_rest.object_api', bucket_id=bucket.id, key=key)
+    # Add a new before request hook which reads the incoming request payload.
+    # This simulates what happens when Sentry's raven-python library whern it
```
ntarocco: `whern` -> `when`

tests/test_views_multipart.py
```diff
@@ -543,3 +543,33 @@ def test_get_listuploads(client, db, bucket, multipart, multipart_url,
             bucket_id=str(bucket.id),
         ) + '?uploads')
         assert res.status_code == expected
+
+
+def test_already_read_input_stream(app, client, db, bucket, admin_user):
```
ntarocco: 👍 👍 👍

invenio_files_rest/views.py
diegodelemos: I am not sure if it makes sense to give an error message on such a low-level detail since the message itself is ambiguous and users won't be able to compose a correct request after reading it...

One could argue that now it is Sentry reading the stream when the content type is `application/json` so we could say "please use `application/octet-stream` instead of `application/json`", but later it could be another library which reads `application/octet-stream` for a different reason and we will be giving a misleading message. Wha do you think? :)

lnielsen: I would create a specific exception and raise, this way it's completely clear in Sentry what the issue is:

```python
#errors.py

class ExhausedStreamError(FilesException):
    pass

# views.py
raise ExhausedStreamError()

Issue #198, opened 2019-02-19
docs: usage documentation


lnielsen:
- REST API
- Storage backends
- Serving files (incl. security)
- Uploading files (incl. JS uploaders and multipart uploading)
- Permissions
- Integrity checking
- Data migration
- Large file uploads
   - Custom Flask class/formparser.
   - Nginx
   - ``MAX_CONTENT_LENGTH``
- Signals (file_downloaded)
- Building your own storage backend.

vladbmihaescu:
DRAFT PR for REST API Usage: https://github.com/topless/invenio-files-rest/pull/2

vladbmihaescu:
[JS Uploaders]: We would like to know in more detail what the scope of this section should be. Additionally, in the code there are a couple of helper methods `ngfileupload_partfactory` and `ngfileupload_partfactory` which I am not sure of how they would work.

PR #210, opened 2019-05-21
Docs usage

vladbmihaescu:
Adds first draft of the usage documentation for the module

(Addresses  #198)


docs/usage/permissions.rst
```diff
@@ -0,0 +1,55 @@
...
+
+
+Access control
+--------------
+
+Invenio-Files-Rest depends on `Invenio-Access
+<https://invenio-access.readthedocs.io>`_ module, to control the files access.
+
+It comes with a default permission factory implementation which can be found
+at :code:`invenio_files_rest.permissions.permission_factory` and can be
```
roksys: ``` :py:data:`invenio_files_rest.permissions.permission_factory ```

docs/usage/permissions.rst
```diff
@@ -0,0 +1,55 @@
...
+customized further, by providing our custom implementation in the relevant
+config variable :code:`FILES_REST_PERMISSION_FACTORY`.
```
roksys: This also could be a link as I mentioned above - ``` :py:data:`invenio_files_rest.config.FILES_REST_PERMISSION_FACTORY` ```

docs/usage/migration.rst
```diff
@@ -0,0 +1,31 @@
...
+
+    Invenio is free software; you can redistribute it and/or modify it
+    under the terms of the MIT License; see LICENSE file for more details.
+
+
+[WIP] Data Migration
+--------------------
+
+:code:`Locations` are used to represent different storage systems and possibly
+different geographical locations. :code:`Buckets` but also :code:`Objects` are
```
ntarocco: I am not convinced by `geographical`, but it might make sense and I miss some history when this was created. We might ask Lars confirmation here.

docs/usage/migration.rst
```diff
@@ -0,0 +1,31 @@
...
+assigned a Location. This architecture provides extra flexibility when we want
+to migrate our data.
+
+When we create a bucket we need to provide a Location, otherwise the default
+one is used.
+
```
ntarocco: Objects are not directly assigned to locations, right? They belongs to a bucket which has a Location.

docs/usage/migration.rst
```diff
@@ -0,0 +1,31 @@
...
+    new storage with the new location. Then we can just bulk update our
+    FileInstance objects pointing to the new bucket.
+
+There is a task :code:`invenio_files_rest.tasks.migrate_file`
+which is used to migrate a file instance to a new location. A new location
+might be in remote system on a different bucket, even on a different storage
+backend. If the process does not complete successfully then the destination
+FileInstance is removed completely. The task also accepts an argument
+:code:`post_fixity_check` which enables the file checksum verification after
+the operation is complete.
```
ntarocco: Giving my version, feel free to adapt/change it, etc...

`
Invenio-Files-REST provides a Celery task (`invenio_files_rest.tasks.migrate_file`) to migrate a existing files from the current location to a new location. Given a `file_id` and the `name` of the new location (which should have been already created), it will:
1. create a new empty FileInstance in the destination location
2. copy the file content in the newly created FileInstance
3. re-link all ObjectVersions pointing to the previous FileInstance to the new one
4. optionally, re-compute the file checksum

This task can be used to migrate all files or a subset of files in case of location change.
`

docs/usage/permissions.rst
```diff
@@ -0,0 +1,55 @@
...
+    Copyright (C) 2015-2019 CERN.
+
+    Invenio is free software; you can redistribute it and/or modify it
+    under the terms of the MIT License; see LICENSE file for more details.
+
+
+Access control
+--------------
+
+Invenio-Files-Rest depends on `Invenio-Access
```
ntarocco: -> Invenio-Files-REST to be consistent everywhere

what about `depends` -> `takes advantage of` or similar?

docs/usage/storage_backends.rst
```diff
@@ -0,0 +1,39 @@
...
+
+In the configuration of the application, we can set the path of the factory that we want to use in order to create a storage instance.
+Each file instance uses the storage factory to create a storage interface for itself.
+
+Build your own Storage Backend
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Advanced topic on how to implement and connect your own storage Backend for
+invenio-files-rest.
+
+In order to use a different storage backend, we just need to subclass the FileStorage class,
```
roksys: ``` :py:data:`invenio_files_rest.storage.FileStorage` ```

docs/usage/upload_files.rst
```diff
@@ -0,0 +1,199 @@
...
+        "tags": {},
+        "checksum": "md5:d7d02c7125bdcdd857eb70cb5f19aecc",
+        "created": "2019-05-16T13:10:22.617714+00:00",
+        "version_id": "7f62676d-0b8e-4d77-9687-8465dc506ca8",
+        "delete_marker": false,
+        "key": "my_file.txt",
+        "size": 14
+    }
+
+Should we want to update the same file with a new version we would use the same command. Automatically, the file will be versioned.
```
roksys: `If we want...`?

docs/usage/permissions.rst
```diff
@@ -0,0 +1,55 @@
...
+<https://invenio-access.readthedocs.io>`_ module, to control the files access.
+
+It comes with a default permission factory implementation which can be found
+at :code:`invenio_files_rest.permissions.permission_factory` and can be
+customized further, by providing our custom implementation in the relevant
+config variable :code:`FILES_REST_PERMISSION_FACTORY`.
+
```
ntarocco: Some suggestions:
`
It provides a default implementation of permissions for several different actions, defined by (`path`), but it also allows to provide your own custom implementation by using the config var <name>
`

docs/usage/upload_files.rst
```diff
@@ -0,0 +1,199 @@
...
+       "bucket":"c896d17b-0e7d-44b3-beba-7e43b0b1a7a4",
+       "completed":false,
+       "part_size":6291456,
+       "key":"my_file.txt",
+       "last_part_number":1,
+       "id":"a85b1cbd-4080-4c81-a95c-b4df5d1b615f",
+       "size":11534336
+    }
+
+Continue uploading parts, by using the PUT command and specifying the id of the upload:
```
roksys: I am not sure if we should use `PUT command` or `PUT request`, since `curl` is a command and `PUT` is a type of request. What do you think?

docs/usage/upload_files.rst
vladbmihaescu: https://ell.stackexchange.com/questions/84695/should-versus-if

invenio_files_rest/__init__.py
```diff
@@ -6,7 +6,158 @@
...
+Invenio-Files-REST guarantees data integrity by calculating the checksum for
+every file and storing it as a property to its :code:`FileInstance`.
+
+Whenever an operation like :code:`merge_multipartobject` the database is rolled
+back to reflect the state of the system before the action.
+
+For the computation of the checksum we can provide, the desired algorithm,
+otherwise :code:`MD5` will be used.
+
+In case like like when we are trying to remove a file, we are expecting
```
roksys: `like like`

docs/usage/serve_files.rst
```diff
@@ -0,0 +1,39 @@
...
+
+Download a specific version of the file:
+
+.. code-block:: console
+
+   $ curl -i http://localhost:5000/files/$B/my_file.txt?versionId=<version_id>
+
+**NOTE:** By default, the file contents are displayed in the browser.
+Should you want to download the file, add the :code:`download` keyword in the query parameters.
+
```
ntarocco: `
Retrieve Files
-------------

Once we have created our bucket and uploaded a file, it is possible to retrieve it with a GET request.
By default, the latest version will be downloaded.

ex

If you want to retrieve a specific version of the file, you can use the versionId query parameter:

 **NOTE:** By default, the file is server with the header 'Content-Disposition': 'inline', meaning that the browser will try to preview it if possible. If you want to trigger a download, you can use the extra query parameter `download`, which will change the 'Content-Disposition' header to 'attachment'
`

docs/usage/storage_backends.rst
```diff
@@ -0,0 +1,39 @@
+..
+    This file is part of Invenio.
+    Copyright (C) 2015-2019 CERN.
+
+    Invenio is free software; you can redistribute it and/or modify it
+    under the terms of the MIT License; see LICENSE file for more details.
+
+Storage Backends
+----------------
+In order to get started we need to setup and configure a storage backend.
```
ntarocco: I would in general first explain concepts. Why do we need to get started with a storage? What is a Storage backend? Why this abstraction?
https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html
https://cloud.google.com/storage/docs/storage-classes

I would mention something like that Invenio provides a default storage that is local storage using standard python lib (PyFSFileStorage). It defines an interface to create your own storage adapter. For example, Invenio-XRootD provides XRootD integration.

docs/usage/upload_files.rst
```diff
@@ -0,0 +1,199 @@
...
+    Copyright (C) 2015-2019 CERN.
+
+    Invenio is free software; you can redistribute it and/or modify it
+    under the terms of the MIT License; see LICENSE file for more details.
+
+
+Creating Buckets
+----------------
+
+In order to upload, modify or delete files, we first need to create a bucket.
```
ntarocco: https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html
https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html

A bucket is a container for objects. Every object is contained in a bucket.
Buckets have a default location (link to the related section) and storage class (link to the related section). 

Each bucket can also define a quota. The size of a bucket is the size of all objects in the bucket (including all versions).

docs/usage/upload_files.rst
```diff
@@ -0,0 +1,199 @@
...
+A bucket can be created by a POST request to the url prefix, i.e. /files.
+The response will contain the unique ID of the bucket.
+A bucket can have one or more tags which store extra metadata for that bucket.
```
ntarocco: Maybe explain what is a tag: extra metadata to "attach" extra information and describe a bucket. It is a loose key/value.

docs/usage/upload_files.rst
```diff
@@ -0,0 +1,199 @@
...
+            "self": "http://localhost:5000/files/cb8d0fa7-2349-484b-89cb-16573d57f09e",
+            "uploads": "http://localhost:5000/files/cb8d0fa7-2349-484b-89cb-16573d57f09e?uploads",
+            "versions": "http://localhost:5000/files/cb8d0fa7-2349-484b-89cb-16573d57f09e?versions"
+        },
+        "created": "2019-05-16T13:07:21.595391+00:00",
+        "quota_size": null,
+        "id": "cb8d0fa7-2349-484b-89cb-16573d57f09e",
+        "size": 0
+    }
+
```
ntarocco: So in general, IMHO, I would first explain the concepts. Imagine I don't know anything about Bucket/ObjectVersion. I need to have an overview of the data model first (we need a location, we need a bucket, then we can upload a file, that is represented by ObjectVersion/FileInstance)
WDYT?

docs/usage/migration.rst
topless: Its not like we came up with it, I saw it here https://github.com/inveniosoftware/invenio-files-rest/blob/4ba19081e738bac8c00e0ddbbc95499ab54ba009/invenio_files_rest/models.py#L26

docs/usage/migration.rst
topless: From what I understood, they actually do! Let's look into that.

docs/usage/upload_files.rst
topless: `PUT` is a request

topless:
> I think we should also update `README.rst` and follow "standard" pattern as in other packages.
> 
> ```
> Short package description
> 
> Features:
> - some feature
> - another feature
> 
> Further documentation is available on ...
> ```
> 
> https://invenio-records-files.readthedocs.io/en/latest/
> https://invenio-records.readthedocs.io/en/latest/

I would agree with you, we have compiled a list of features in the beginning of the usage section. We can enhance and move that part to the README as you suggested.

invenio_files_rest/__init__.py
```diff
@@ -6,7 +6,730 @@
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
-"""Files download/upload REST API similar to S3 for Invenio."""
+
+r"""REST API for Files similar to S3 for Invenio.
+
+Invenio-Files-REST is a core component of Invenio which provides configurable
+REST APIs for uploading, serving, downloading and deleting files. It works in
```
diegodelemos:
```
It works as a standalone module or in combination with `Invenio-Records <https://invenio-records.readthedocs.io>`_
through the `Invenio-Records-Files <https://invenio-records-files.readthedocs.io>`_ integration.
```

What do you think? And it would link to this https://github.com/inveniosoftware/invenio-records-files/pull/79/files#diff-f2377f695b4ac4cabc53efd87274b35eR184.

Feel free to rephrase if you like it.

invenio_files_rest/__init__.py
```diff
@@ -6,7 +6,730 @@
...
+but also as a standalone module.
+
+The module can be configured with different storage backends, and provides
+features such as:
+
+- A robust REST API.
+- Configurable storage backends with the ability to build your very own.
+- Highly customizable permission mechanism for access-control.
+- Secure file handling.
+- Integrity check mechanism.
```
diegodelemos: `Integrity checking mechanism`?

invenio_files_rest/__init__.py
```diff
@@ -6,7 +6,746 @@
...
+
+Now let's create a bucket:
+
+>>> res = app.test_client().post('/files')
+
+And see the response containing the id of the bucket:
+
+>>> pprint(json.loads(res.get_data().decode("utf-8")))
+
+Data Model
```
diegodelemos: Maybe this clarifications could be treated as the overview in Invenio-Access (see [here](https://invenio-access.readthedocs.io/en/latest/overview.html)), what do you think?

docs/overview.rst
```diff
@@ -0,0 +1,67 @@
...
+    Copyright (C) 2015-2019 CERN.
+
+    Invenio is free software; you can redistribute it and/or modify it
+    under the terms of the MIT License; see LICENSE file for more details.
+
+
+Overview
+========
+
+In order to upload files, a Location needs to be defined first.
```
ntarocco: I would add an extra gentle introduction paragraph, something like (please improve!):

> Invenio-Files-REST is a files storage module. You can store and retrieve files in a similar way to Amazon S3 APIs.
> 
> To understand how to safely manage files in Invenio, this guide will introduce you to the key concepts and terminology that you need to understand.

I will, IMHO, but please let's understand together what is best, start from "the end", files: this is because if I am  new to these concepts, I don't know/understand why I need locations and buckets.
I would try to answer this question: I have a file, why would I need a location/bucket/object version?
I would also do **not** focus on upload/download only, because this if for HTTP. We can use the module programmatically.

The text below should be improved... but you should get my point.

> Files are represented by ``FileInstance``,  which describes the path (or location???) the file on disk. Each file belongs to a ``Bucket``, a container of files. Buckets are important to organize your storage or, for example, to define quotas. Since a file can have a complex set of metadata, for example name, type, size, etc., an ``ObjectVersion`` is associated to each file.

And then, maybe start the detailed description of each object.


> Locations
> ----------
> 
> A Location is ...

invenio_files_rest/__init__.py
```diff
@@ -6,7 +6,743 @@
...
+
+- A robust REST API.
+- Configurable storage backends with the ability to build your very own.
+- Highly customizable permission mechanism for access-control.
+- Secure file handling.
+- Integrity check mechanism.
+- Supports for large file uploads and multipart upload.
+- Signals for system events.
+
+The REST API follows best practices and supports, e.g.:
```
okraskaj: Because of the comma after supports it sounds like supports was meant to be a plural form of a noun- and for me this doesn't really fit in the sentence. I would either remove the comma so it's clear that supports is a verb, or even better phrase it somehow along the lines of "... and provides following features:"

invenio_files_rest/__init__.py
```diff
@@ -6,7 +6,744 @@
...
+GET
+^^^
+
+Return list of files in bucket:
+
+.. code-block:: console
+
+    GET /files/<bucket_id>/
+
+Returns list of file versions:
```
okraskaj: The form of the verb is inconsistent- I would stay with "Return" also for the next 2 occurrences.

invenio_files_rest/__init__.py
```diff
@@ -6,7 +6,744 @@
...
+
+   $ curl -i http://localhost:5000/files/$B/my_file.txt?download
+
+
+Security
+^^^^^^^^
+It is very easy to be exposed to Cross-Site Scripting (XSS) attacks if you
+serve user uploaded files. Here are some recommendations:
+
+1. Serve user uploaded files from a separate domain (not a subdomain). This
```
okraskaj: All the points are rendered without any line brake- maybe we should add newlines between them

invenio_files_rest/__init__.py
```diff
@@ -6,7 +6,788 @@
...
+.. code-block:: python
+
+    @need_permissions(
+        lambda self, bucket, versions: bucket,
+        'bucket-read',
+    )
+    def my_function():
+        pass
+
+.. note::
```
okraskaj: Probably a note to self.

invenio_files_rest/__init__.py
```diff
@@ -6,7 +6,788 @@
...
+it's possible to proceed with the removal of the actual file. By binding the
+actions together, in case of database operation error, you can prevent the
+system ending up with orphan files. Some of the tasks support :code:`silent`
+as an argument to prevent propagation of the errors.
+
+There is also a predefined task :code:`verify_checksum` which can be configured
+to run periodically (default is every 30 days) which iterates all files in our
+storage and validates their checksum.
+
+>>> "Get the checksum of an uploaded file for demo"
```
okraskaj: @topless I think this is a little bit out of context. Maybe two lines of python code that you mentioned would be better.

docs/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+allowing some operations to be performed more efficiently, such as snapshots
+without duplicating files or migrating data.
+Just as in a computer files are contained inside folders, each :code:`Object` has
+to be contained in a :code:`Bucket`. The bucket is identified by a unique ID,
+assigned automatically at creation. A :code:`Bucket` is created by default in the
+default :code:`Location`, however that can be changed such that when creating a
+:code:`Bucket`, a particular :code:`Location` for it can be specified. The
+:code:`Bucket` can also have a maximum quota assigned to it, and an important
+point to note is that the :code:`Objects` inside it do not necessarily have to
+be located in the same :code:`Location`. The :code:`Location` can be used to
```
egabancho: > and an important point to note is that the :code:`Objects` inside it do not necessarily have to
be located in the same :code:`Location`

Hey! Are you sure about that? The `Bucket` is the one that has the `Location` as a thing handled by Invenio-Files-Rest, I understand that you can set the "physical location" of a file using [`ObjectVersion.set_location`](https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/models.py#L1006) but I think this is used on the special case where the file is outside "our file system", i.e. linked.
I might be wrong ☺️

docs/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+that one may use to store extra information.
+The tags are identified uniquely within a bucket.
+
+
+Objects
+-------
+Objects are an abstraction of a file, and are uniquely identified within
+a bucket by string keys, i.e. the file name.
+
+
```
egabancho: Object Tags? ☺️

docs/overview.rst
```diff
@@ -0,0 +1,97 @@
...
+--------
+Locations are used to represent different storage systems and/or geographical
+locations.
+
+
+Storage
+-------
+Storage classes are useful for defining the type of storage an object is
+located on (e.g. offline/online), so that the system knowns if it can serve
+the file and/or what is the reliability.
```
ntarocco: I think we need to describe better why we need a Location and a Storage, let's see together.

ntarocco:
@topless @vladbmihaescu @Glignos honestly, nice job! I think the overview is a great starting point for users.
I managed to review the Overview, README and started with the Usage.
I have added a commit just to re-arrange the order of concepts in the Overview.

For the REST API, available methods, etc... I think it should be split because it is mixed between location, buckets, files. I think it would better if as a user I can see the available REST APIs for:
- Location: verbs and explanations
- Buckets: verbs and explanations
- Files: verbs and explanations

Tomorrow I will continue will the rest.

ntarocco:
@topless actually I cannot push... did you uncheck the [allow edits from maintainers](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork)?

ntarocco:
New one: https://github.com/inveniosoftware/invenio-files-rest/pull/227

PR #216, opened 2019-07-25
global: marshmallow 2 and 3 compatibility

kprzerwa:


invenio_files_rest/serializer.py
```diff
@@ -18,6 +19,44 @@
...
+
+        def data_errors(self):
+            warnings.warn(
+                "Schema().dump().errors attribute is deprecated "
+                "in marshmallow v3.x",
+                category=DeprecationWarning, stacklevel=2)
+            return result
+
+        setattr(cls, 'data', property(data))
+        setattr(cls, 'errors', property(data_errors))
```
kprzerwa: after a thought, I would remove this and let it fail on errors, or I would raise an exception, but I will wait for a feedback

setup.py
```diff
@@ -63,11 +63,14 @@
     'Flask>=0.11.1',
     'fs>=0.5.4,<2.0',
     'invenio-rest[cors]>=1.0.0',
-    'marshmallow>=2.15.1,<3.0.0',
+    'marshmallow>=2.15.1,<3.0.0;python_version<"3.0.0"',
```
lnielsen: Please see https://github.com/inveniosoftware/troubleshooting/issues/1

invenio_files_rest/serializer.py
```diff
@@ -18,6 +19,44 @@
...
 
 
+class MarshalResultBase(object):
+    """Wrapper class to provide backward compatibility with marshmallow 2."""
+
+    def __new__(cls, result, errors=None, *args, **kwargs):
+        """Instanciate Marshal Result class."""
+        def data(self):
+            warnings.warn(
+                "Schema().dump().data attribute is deprecated "
```
lnielsen: Will this issue the warning on Marshmallow 2? 

I think the warning should only be issued when v3 is installed to warn users that their schema will break at some point.

I guess this class is also useful for other modules. Perhaps we need to find another home for it?

lnielsen:
Will you update once invenosoftware/invenio-rest#94 is done? I think it still makes sense to include marshmallow as a devel test.

PR #221, opened 2019-11-13
serializer: make serializer methods more versatile

Glignos:


invenio_files_rest/serializer.py
```diff
@@ -52,7 +52,7 @@ def dump_links(self, o):
         }
 
 
-class ObjectVersionSchema(BaseSchema):
+class BasicObjectVersionSchema(BaseSchema):
```
ntarocco: I would use the work `Base` instead

invenio_files_rest/serializer.py
```diff
@@ -268,7 +269,8 @@ def _whitespace_waiting():
 
 
 def json_serializer(data=None, code=200, headers=None, context=None,
-                    etag=None, task_result=None):
+                    etag=None, task_result=None,
+                    serializer_mapping=serializer_mapping, view_name={}):
```
ntarocco: I guess that if we go for the param `view_name`, and not `kwargs`, I would put a string and not a dict...

PR #222, opened 2019-11-15
global: bump invenio version for marshmallow support

ntarocco:


invenio_files_rest/serializer.py
```diff
@@ -107,7 +107,7 @@ def wrap(self, data, many):
             data = {'contents': data}
             bucket = self.context.get('bucket')
             if bucket:
-                data.update(BucketSchema().dump(bucket).data)
+                data.update(BucketSchema().dump(bucket))
```
topless: This looks like a marshmallow change from 2 to 3, maybe for that module rerun the tests inveniosoftware/invenio#3967

PR #226, opened 2019-11-20
signals: add signal for deletion and upload of files.

Glignos:


ChiaraBi:
Should we add some tests? Otherwise LGTM :)

invenio_files_rest/signals.py
```diff
@@ -19,3 +19,16 @@
...
+
+file_uploaded = _signals.signal('file-uploaded')
+"""File uploaded signal.
+
+Sent when a file is uploaded.
+"""
+
+
+file_deleted = _signals.signal('file-deleted')
+"""File delete signal.
```
ntarocco: fix me: `File deleted signal.`

ntarocco:
> Should we add some tests? Otherwise LGTM :)

@ChiaraBi they will come in the next PR! Thanks!

Issue #234, opened 2020-01-23
cli: list locations


ppanero:
Related to https://github.com/inveniosoftware/invenio-files-rest/issues/233 In order to remove a location I might first need to list them.

Issue #232, opened 2020-01-23
new default location should remove previous "default" ones.


ppanero:
If a location has been created with `--default` and a new one is created the same way, both have the default flag to `true` in the DB. This ends up in a `None` location in the code due to many being return.

The `-default` should set to false previous default locations.

PR #235, opened 2020-01-23
list locations

wgresshoff:
In order to set a default location I need to know which are existing.

Closes https://github.com/inveniosoftware/invenio-files-rest/issues/234
Closes https://github.com/inveniosoftware/invenio-files-rest/issues/232

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,12 @@
 Changes
 =======
 
+Version 1.0.6 (released 2019-11-22)
+
+- Bump version and add to installation requirements invenio-celery
+- Add documentation of module usage
+- Remove storage_class parameter from Bucket create when POST to Location resource
+
 Version 1.0.5 (released 2019-11-21)
 
 - Add signals for deletion and upload of files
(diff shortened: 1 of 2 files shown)


ppanero:
As discussed IRL, the "set default" functionality will be generic so the create location will also make use of it when calling with ``--default``

invenio_files_rest/cli.py
```diff
@@ -67,4 +67,33 @@ def location(name, uri, default):
...
+    click.secho('-' * 169)
+    for location in locations:
+        click.secho(str(location), fg='green')
+
+@files.command()
+@click.argument('name')
+@with_appcontext
+def default_location(name):
+    """Set default location."""
+    from .models import Location
```
ppanero: Imports should be on the beginning on the file.

wgresshoff:
Just added fix for https://github.com/inveniosoftware/invenio-files-rest/issues/232

ppanero:
Slight refactoring to fixt tests. (@wgresshoff sorry I had to take out the pretty print it is a nice addition but it requires a bigger refactor along invenio).


**Warning** it is not backwards compatible:

- CLI `location` command is now a group to contain its related commands `create`, `list` and `set_default`.
- `Location.get_by_name()` returns now `one_or_none` instead of `one` and raising an exception if it does not exist.

ppanero:
Continued in https://github.com/inveniosoftware/invenio-files-rest/pull/236

PR #236, opened 2020-01-30
cli: allow listing and setting default of locations

### LINKED ISSUE #232: new default location should remove previous "default" ones.

### LINKED ISSUE #234: cli: list locations

ppanero:
Changes:
* Allows listing locations
* Allows setting a location as defautl (by name)
* Fixes a bug: When a second location was created with ``--default`` there were two defaults in DB and therefore failing the uploading of files.

Duplicated of https://github.com/inveniosoftware/invenio-files-rest/pull/235 Since the PR was from master branch I could not force push to squash commits.

Slight refactoring to fixt tests. (@wgresshoff sorry I had to take out the pretty print it is a nice addition but it requires a bigger refactor along invenio).

**Warning** it is not backwards compatible:

- CLI `location` command is now a group to contain its related commands `create`, `list` and `set_default`.
- `Location.get_by_name()` returns now `one_or_none` instead of `one` and raising an exception if it does not exist.

**Requires**

- Update cookiecutter [steup script](https://github.com/inveniosoftware/cookiecutter-invenio-instance/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/scripts/setup#L12)

Closes #234
Closes #232


invenio_files_rest/cli.py
```diff
@@ -56,15 +59,58 @@ def cp(source, bucket, checksum, key_prefix):
...
+def create(name, uri, default):
+    """Create a new location."""
+    if default:
+        _unset_default_location()
     location = Location(name=name, uri=uri, default=default)
     db.session.add(location)
     db.session.commit()
-    click.secho(str(location), fg='green')
+    click.secho('Location {} {} as default {} created'.format(
+        location.name, location.uri, str(location.default)), fg='green')
```
fenekku: Can we make this operation idempotent instead? i.e. make `create` (can rename) ensure that a `Location` with the specified name/uri/default exists (only creates it if it doesn't)? This would make its usage on the command line much smoother: it can be repeatedly run without issue. Right now, this will raise an exception if a Location already exists, even though at the end of the day the goal is just to have this entry in the DB.

setup.py
```diff
@@ -65,6 +65,7 @@
     'invenio-celery>=1.1.1',
     'invenio-rest[cors]>=1.1.2',
     'simplejson>=3.0.0',
+    'six>=1.12.0',
```
fenekku: now that we are on Python 3, we should not need six anymore right?

ppanero: Since this module was there before we keep support for Python 2.x for a bit (AFAIK).

invenio_files_rest/cli.py
ppanero: I see your point. However, since the PK is only the `name` is unique the only editable fields are `uri` and `default`. I only see the use case to change the default, if you need a new `uri` you might as well just create a new location. 

IMHO deleting/removing or even editing a location might cause undesired effects if it already contains files. 

That said I agree on allowing the command be run multiple times without exception (i.e. catching it and showing a red message like "location name already exists".

@fenekku  WDYT?

fenekku: Yes that would work. Something like: if name / uri already exists, just set default if explicit default, else show message (probably yellow) about location already existing. If the name doesn't exit, create.

setup.py
fenekku: ah ok!

invenio_files_rest/cli.py
ppanero: Fixed.

It now shows a yellow message if the name exists. However, following the same logic than above I opted by not setting it default, since for that there is a command (and again it will require explicit an action by the user).

ppanero:
Fails due to broken imports from Werkzeug (coming from bablex).

setup.py
fenekku: *I already gave my approval and should have provided those yesterday, so feel free to ignore my comments. I try to be consistent and want to get those remarks out there*: 

if we are keeping support for python 2 for a bit, we should keep testing against it in Travis. We should only remove the Python 2 tests, when we remove the support for it I think.

tests/test_cli.py
```diff
@@ -26,11 +26,67 @@ def test_simple_workflow(app, db, tmpdir):
...
+    assert 0 == result.exit_code
+
+    created_locations = result.output.split('\n')
+    # tmp is default
+    assert "tmp" in created_locations[0]
+    assert "as default True" in created_locations[0]
+    # aux is not default
+    assert "aux" in created_locations[1]
+    assert "as default False" in created_locations[1]
+
```
fenekku: *I already gave my approval and should have provided those yesterday, so feel free to ignore my comments. I try to be consistent and want to get those remarks out there*: 

The already existing tests here were testing against the cli output, so I see the appeal in following in their footsteps. However, given they interact with the DB, testing what's in the DB is probably more accurate and valuable than what is printed out. But I am not sure where the line is for tests that run fast, are valuable and are convenient to write. Maybe what we have here is a good enough comprise...

setup.py
ppanero: This module (invenio-files-rest) is already released (is not like RDM ones) so this one keeps support for Python 2 (as you can see in travis, it keeps testing those: https://github.com/inveniosoftware/invenio-files-rest/blob/master/.travis.yml#L57)

The new RDM modules we are developing are alpha/not stable yet, and since no one should be relying on them until June we only need to support Python 3.

tests/test_cli.py
ppanero: I kind of agree with you. Due to time constraints (and the fact that it works -  tested manually) we I'd go with it and open an issue. Since the test against db should be done not only for my new commands but for all the CLI ones (buckets, etc.)

invenio_files_rest/models.py
```diff
@@ -286,7 +286,7 @@ def get_by_name(cls, name):
         """Fetch a specific location object."""
         return cls.query.filter_by(
             name=name,
-        ).one()
+        ).one_or_none()
```
lnielsen: What's the reason behind this change? I mean, it will go from throwing an exception to returning None, so any code relying on an exception will break

ppanero: It's kind of two folded:

- Consistency: All the other `get` operations in the models.py return a `one_or_none`. We didn't understood why this one exactly was returning just `one`.
- Makes the CLI code simpler. We just check for `None` istead of `try/except`.

Nonetheless, I understand that this would make it a minor version. However, since the CLI has breaking changes two (`location` becomes a group not a command). It would need it anyway, no?

ppanero:
@lnielsen 
> Are the broken tests related to the Werkzeug issues?

Yes, coming from:
- flask-babelex
- flask-wtf
- invenio-admin 

I bumped the minor version in all. However, `flask-wtf` is bumped a full minor. I checked the released notes and some method signatures have changed... Tests do not break, impact on the community - this PR will be a minor bump anyway - No?

It still fails on `release` due to a version conflict: on `Flask-Login`  because `invenio-accounts` has `<5.0.0` and `invenio-admin` `>5.0.05

setup.py
fenekku: My point was that Travis testing of Python 2 is removed *in this PR* . See at the top...

ppanero: Yeah, you are right. Mistake, talked to Lars. We do not "test" it anymore (e.g. if it would break we wouldn't fix it) however removing `six` has more implications since it is used widely across the code base of the module (and it will work anyway with Python 3).

Removing it will be a separate issue/task at some point.

ppanero:
@lnielsen tests are passing :)

setup.py
fenekku: I see... I was further confused by the fact that `six` is *added* here, although it's already used... I made the issue for switching to Python 3 completely: https://github.com/inveniosoftware/invenio-files-rest/issues/239 .

fenekku:
All good on my side. I am not authorized to merge though; someone else will have to do it.

PR #242, opened 2020-02-24
Release v1.1.1

fenekku:
Fix backwards compatibility of `invenio files location` (make that still possible and default to `create` command which was the functionality before).

### DIFF
--- a/invenio_files_rest/cli.py
+++ b/invenio_files_rest/cli.py
@@ -12,6 +12,7 @@
 from __future__ import absolute_import, print_function
 
 import click
+from click_default_group import DefaultGroup
 from flask.cli import with_appcontext
 from invenio_db import db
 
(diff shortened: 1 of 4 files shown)


ppanero:
AFAIK, since we passed from 1.0 to 1.1 it was allowed to break that, why is it absolutely needed to support back? there is no other change.

fenekku:
This is not motivated by a theoretical best practice but really by a practical advantage.

This eliminates the need for _us or any other package_ to change code calling invenio-files-rest cli commands. In a practical sense it saves work across the board: work to change, work to figure out why the command stopped working for a feature bump. 
 
... Invenio doesn't strictly follow semantic versioning I know, but it sorts of advertises itself as though it does: it would be more stable if it did more, doing so saves work, the third-party libraries that we use mostly do and finally it would help with our own version specifier woes. I argue we should attempt to follow it for the changes we make. I will bring this point at the retrospective. Unless absolutely necessary, interfaces should be made backward compatible, because it saves us work.

lnielsen:
> Unless absolutely necessary, interfaces should be made backward compatible, because it saves us work.

I definitely agree :-)

PR #245, opened 2020-03-05
views: stream optionally files through NGINX

Glignos:


invenio_files_rest/config.py
```diff
@@ -127,3 +129,10 @@
 
 FILES_REST_FILE_TAGS_HEADER = 'X-Invenio-File-Tags'
 """Header for updating file tags."""
+
+FILES_REST_OFFLOAD_STREAMING_EXTERNALLY = False
```
ntarocco: I would say that in general this feature is known as `X-Sendfile`, but it would be nice if you can check this.
`X-Accel-Redirect` is also something only for Nginx? I guess that we would like to provide something more generic, maybe with a default implementation for Nginx. What do you think?

invenio_files_rest/config.py
```diff
@@ -127,3 +129,10 @@
...
+"""Use the X-Accel-Redirect header to stream the file through NGINX"""
+
+FILES_REST_STREAMING_REDIRECT_RESPONSE_FUNC =  \
```
ntarocco: Continuing my previous comment, what about if we for something similar to:
```
FILES_REST_XSENDFILE_ENABLED = False

FILES_REST_XSENDFILE_RESPONSE_FUNC = ...
```
not sure about the second though, we should check in Invenio what we are using....

invenio_files_rest/views.py
```diff
@@ -814,6 +814,10 @@ def get(self, bucket=None, key=None, version_id=None, upload_id=None,
             return self.multipart_listparts(bucket, key, upload_id)
         else:
             obj = self.get_object(bucket, key, version_id)
+            if current_app.config['FILES_REST_OFFLOAD_STREAMING_EXTERNALLY']:
+                response_constructor = current_app.config[
+                    'FILES_REST_STREAMING_REDIRECT_RESPONSE_FUNC']
+                return response_constructor(obj)
```
ntarocco: Silly question: if we pass "only" the `obj`, and inside we do `obj.bucket`, does SQLAlchemy perform another db query? If yes, maybe we want to pass more params...

tests/test_views_objectversion.py
```diff
@@ -82,6 +82,29 @@ def test_get(client, headers, bucket, objects, permissions):
                 assert resp.get_etag()[0] == obj.file.checksum
 
 
+def test_get_with_redirect(
```
ntarocco: Same about naming here, maybe something more related to `X-Sendfile`?

invenio_files_rest/helpers.py
```diff
@@ -291,3 +292,13 @@ def create_file(key, path):
                 assert filename.startswith(source)
                 parts = [p for p in filename[len(source):].split(os.sep) if p]
                 yield create_file('/'.join(parts), os.path.join(root, name))
+
+
+def create_file_streaming_redirect_response(obj):
```
ntarocco: What about if someone has to create a signed url for the final file, with maybe a token per user? Do we have enough params? Does it make sense to do a quick search how other people need this?
E.g. first link on Google:
https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/s3-example-presigned-urls.html

ntarocco: As discussed in IRL, who is implementing this function should already have enough objects to access what needed.

invenio_files_rest/helpers.py
```diff
@@ -291,3 +292,13 @@ def create_file(key, path):
...
+    """Redirect response generating function."""
+    response = make_response()
+    redirect_url_base = '/secure_files/'
```
ntarocco: I would name this, as default, differently, because someone could think that by default it is `insecure` :)
I would also name the function something with Nginx...what about something like `create_response_nginx_xsendfile` or similar?

invenio_files_rest/helpers.py
```diff
@@ -291,3 +292,13 @@ def create_file(key, path):
...
+    redirect_url_key = urlsplit(obj.file.uri).path
```
ntarocco: just double check when the `obj` has a delete marker, if everything is ok, otherwise it might fail doing `obj.file`.

invenio_files_rest/helpers.py
```diff
@@ -291,3 +292,15 @@ def create_file(key, path):
                 assert filename.startswith(source)
                 parts = [p for p in filename[len(source):].split(os.sep) if p]
                 yield create_file('/'.join(parts), os.path.join(root, name))
+
+
+def create_file_streaming_redirect_response(obj):
+    """Redirect response generating function."""
+    if obj.deleted:
+        return abort(404)
```
ntarocco: I am sorry to come back on this @Glignos :(

I think that right now when a file is deleted the answer is not a 404, but a 200 with `delete_marker: true`.
See here: https://invenio-files-rest.readthedocs.io/en/latest/usage.html#uploading-files

I am not 100%, are you able to test it?
Maybe create a test for this new `if`?

Glignos: So, in this case, the normal behavior would be that the [get_object](https://github.com/inveniosoftware/invenio-files-rest/blob/a7ece14c644a728597db77075feb15e65f88c5e2/invenio_files_rest/views.py#L547) would abort and return a `404` since the [ObjectVersion.get](https://github.com/inveniosoftware/invenio-files-rest/blob/a7ece14c644a728597db77075feb15e65f88c5e2/invenio_files_rest/models.py#L1172) will return `None` since the filters will reject the deleted marker and the previous versions.
Although this still can be removed since it is redundant, since if the object is deleted the request will be aborted beforehand when it is fetched.
Added also the tests since indeed they were missing.

setup.py
```diff
@@ -62,13 +62,13 @@
...
     'invenio-base>=1.2.2',
+    'invenio-accounts>=1.2.0',
     'fs>=0.5.4,<2.0',
     'invenio-celery>=1.2.0',
     'invenio-rest[cors]>=1.2.0',
     'simplejson>=3.0.0',
-    'six>=1.12.0',
-    'SQLAlchemy-Utils>=0.31.0',
     'WTForms>=2.0',
+    'SQLAlchemy-Utils<0.36,>=0.33.1',
```
ppanero, line 71: minor: Order lower to upper bound.

```suggestion
    'SQLAlchemy-Utils>=0.33.1,<0.36',
```

setup.py
```diff
@@ -62,13 +62,13 @@
     'Flask-Login>=0.3.2,<0.5.0',
     'Flask-WTF>=0.14.3',
     'invenio-base>=1.2.2',
+    'invenio-accounts>=1.2.0',
```
ppanero, line 65: minor: 1.2.0 is a broken build, maybe 1.2.1 is a better candidate, although functionality wise, might not change anything.
```suggestion
    'invenio-accounts>=1.2.1',
```

PR #253, opened 2020-05-14
Remove example app

### LINKED ISSUE #250: Document the use of `FILES_REST_XSENDFILE_ENABLED`
[Related PR](https://github.com/inveniosoftware/invenio-files-rest/pull/245/files)

Glignos:
Closes #250


invenio_files_rest/__init__.py
```diff
@@ -386,6 +387,14 @@
 Be aware that there are security implications to take into account when
 serving files. See the :ref:`usage-security` for more information.
 
+Invenio-Files-Rest provides also the functionality to serve your files directly
+from your external storage. This is achieved by attaching the
+`X-Accel-Redirect` header to the response, which will then be redirected by
```
diegodelemos, line 392: Comment: is there any good resource we can point to that documents this header? If that exists I would appreciate to have a link and learn it.

invenio_files_rest/__init__.py
```diff
@@ -386,6 +387,14 @@
...
+your Web Proxy (e.g. NGINX, Apache) to your external storage, finally streaming
+the file directly to the user.
+To use this you will need to configure your Web Proxy accordingly and then
```
diegodelemos, line 395: Comment: what do you think of:
```suggestion
To use this feature you will need to configure your Web Proxy accordingly and then
```

PR #258, opened 2020-10-15
Bump the isort test dependency version

alexdutton:
This repo uses requirements-builder to test against both minimum and latest versions of dependencies. However, isort 4.x sorts imports inconsistently compared to isort 5.x, which means CI can't currently pass.

Bumping the version forces all test runs to use the same version of isort, until later versions are released.


lnielsen:
Superseded

PR #273, opened 2021-10-26
models: add a fix to the sync method

zzacharo:


ntarocco:
@zzacharo I found another issues with the signals (change of method signature?), we have to change:
`file_downloaded.send(obj)` to `file_downloaded.send(obj=obj)`
and check the other signals too.

PR #280, opened 2022-04-27
Add task for cleaning up orphan files

max-moser:
This PR adds a background task that deletes orphaned files from DB and disk (i.e. `FileInstance`s that don't have any `ObjectVersion`s linked with them anymore).
It is a slightly adapted version of our custom CLI command: https://gitlab.tuwien.ac.at/fairdata/invenio-utilities-tuw/-/blob/master/invenio_utilities_tuw/cli/files.py#L240

**Note**: The CLI command has been working well enough to clean up unreferenced files of deleted drafts, but I found that the `force` flag has to be provided because the storage is marked as not `writable`.

I'll still add tests for the new task, and then mark this as ready for review.

### DIFF
--- a/invenio_files_rest/tasks.py
+++ b/invenio_files_rest/tasks.py
@@ -213,12 +213,14 @@ def migrate_file(src_id, location_name, post_fixity_check=False):
 
 
 @shared_task(ignore_result=True)
-def remove_file_data(file_id, silent=True):
+def remove_file_data(file_id, silent=True, force=False):
     """Remove file instance and associated data.
 
     :param file_id: The :class:`invenio_files_rest.models.FileInstance` ID.
     :param silent: It stops propagation of a possible raised IntegrityError
         exception. (Default: ``True``)
+    :param force: Whether to delete the file even if the file instance is not
+        marked as writable.
     :raises sqlalchemy.exc.IntegrityError: Raised if the database removal goes
         wrong and silent is set to ``False``.
     """
@@ -226,7 +228,7 @@ def remove_file_data(file_id, silent=True):
         # First remove FileInstance from database and commit transaction to
         # ensure integrity constraints are checked and enforced.
         f = FileInstance.get(file_id)
-        if not f.writable:
+        if not f.writable and not force:
             return
         f.delete()
         db.session.commit()
@@ -279,3 +281,54 @@ def remove_expired_multipartobjects():
 
     for fid in file_ids:
         remove_file_data.delay(fid)
+
+
+@shared_task(ignore_result=True)
+def clear_orphaned_files(force_delete_check=lambda file_instance: False, limit=1000):
+    """Delete orphaned files from DB and storage.
+
+    .. note::
+
+        Orphan files are files
+        (:class:`invenio_files_rest.models.FileInstance` objects and their
+        on-disk counterparts) that do not have any
+        :class:`invenio_files_rest.models.ObjectVersion` objects associated
+        with them (anymore).
+
+    The celery beat configuration for scheduling this task may set values for
+    this task's parameters:
+
+    .. code-block:: python
+
+        "clear-orphan-files": {
+            "task": "invenio_files_rest.tasks.clear_orphaned_files",
+            "schedule": 60 * 60 * 24,
+            "kwargs": {
+                "force_delete_check": lambda file: False,
+                "limit": 500,
+            }
+        }
+
+    :param force_delete_check: A function to be called on each orphan file instance
+                               to check if its deletion should be forced (bypass the
+                               check of its ``writable`` flag).
+                               For example, this function can be used to force-delete
+                               files only if they are located on the local file system.
+                               Signature: The function should accept a
+                               :class:`invenio_files_rest.models.FileInstance` object
+                               and return a boolean value.
+                               Default: Never force-delete any orphan files
+                               (``lambda file_instance: False``).
+    :param limit: Limit for the number of orphan files considered for deletion in each
+                  task execution (and thus the number of generated celery tasks).
+                  A value of zero (0) or lower disables the limit.
+    """
+    # with the tilde (~) operator, we get all file instances that *don't*
+    # have any associated object versions
+    query = FileInstance.query.filter(~FileInstance.objects.any())
+    if limit > 0:
+        query = query.limit(limit)
+
+    for orphan in query:
+        # note: the ID is cast to str because serialization of UUID objects can fail
+        remove_file_data.delay(str(orphan.id), force=force_delete_check(orphan))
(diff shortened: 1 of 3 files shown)


invenio_files_rest/tasks.py
```diff
@@ -264,3 +264,40 @@ def remove_expired_multipartobjects():
...
+        (:class:`invenio_files_rest.models.FileInstance`s and their on-disk
+        counterparts) that do not have any
+        :class:`invenio_files_rest.models.ObjectVersion`s associated with them
+        (anymore).
+    """
+    # with the tilde (~) operator, we get all file instances that *don't*
+    # have any associated object versions
+    orphans = FileInstance.query.filter(~FileInstance.objects.any()).all()
+    for orphan in orphans():
+        file_loc = orphan.uri
```
lnielsen, line 283: I would probably just send off a ``remove_file_data`` task:

``remove_file_data.delay(orphan.id)``

E.g. this ensure each files is removed in the most safe way. E.g. it will remove the file instance, commit to the database, and then remove the file. This way we're never left with a file instance in the database that does not have a corresponding file (except if someone from outside decided to remove it).

max-moser: **that** is also the task that inspired the CLI command!

invenio_files_rest/tasks.py
```diff
@@ -195,20 +195,22 @@ def migrate_file(src_id, location_name, post_fixity_check=False):
...
+        marked as writable.
     :raises sqlalchemy.exc.IntegrityError: Raised if the database removal goes
         wrong and silent is set to ``False``.
     """
     try:
         # First remove FileInstance from database and commit transaction to
         # ensure integrity constraints are checked and enforced.
         f = FileInstance.get(file_id)
-        if not f.writable:
+        if not f.writable and not force:
```
max-moser, line 213: InvenioRDM marks files as not writable, e.g. here's two files that i've uploaded for a draft, which i deleted afterwards:

```ipython
In [1]: from invenio_files_rest.models import ObjectVersion, FileInstance, Bucket

In [2]: orphans = FileInstance.query.filter(~FileInstance.objects.any()).all()

In [3]: orphans
Out[3]:
[<FileInstance 4d22743a-282b-4f91-956e-f8a3fc632403>,
 <FileInstance 3f3e1e10-918b-4070-9a2f-e61de0205cc1>]

In [4]: orphans[0].writable
Out[4]: False

In [5]: orphans[1].writable
Out[5]: False
```

max-moser:
updated the PR, tested it out in a local `my-site` and everything works
the changes necessary to make the tests run are contained in https://github.com/inveniosoftware/invenio-files-rest/pull/282

invenio_files_rest/config.py
```diff
@@ -135,3 +141,17 @@
 
 FILES_REST_XSENDFILE_RESPONSE_FUNC = create_file_streaming_redirect_response
 """Function for the creation of a file streaming redirect response."""
+
+FILES_REST_FORCE_DELETE_ORPHAN_FILE = never_force_delete
```
ntarocco, line 145: Super minor:
```suggestion
FILES_REST_FORCE_DELETE_ORPHAN_FUNC = never_force_delete
```

invenio_files_rest/tasks.py
ntarocco: ~~When does InvenioRDM mark files as not writable? Always? Or only when the draft is deleted or marked as deleted?~~
I re-read your explanations, now I have understood. Do you know if there are other cases?

invenio_files_rest/tasks.py
```diff
@@ -279,3 +281,30 @@ def remove_expired_multipartobjects():
...
+        on-disk counterparts) that do not have any
+        :class:`invenio_files_rest.models.ObjectVersion` objects associated
+        with them (anymore).
+
+    :param force: Whether to delete the file even if the file instance is not
+        marked as writable.
+    """
+    # with the tilde (~) operator, we get all file instances that *don't*
+    # have any associated object versions
+    orphans = FileInstance.query.filter(~FileInstance.objects.any()).all()
```
ntarocco, line 301: This query might return high number of results, if not often run or the first time? Or do you want it will always have low cardinality?
It might create high number of celery task.

invenio_files_rest/tasks.py
max-moser: From the top of my head I can't remember any times when InvenioRDM would've set the files to `writable=True`... Gotta check through the codepaths again, to see if it's possible to change that behaviour.

Though likely already running instances would have (at least some, if not all) files "non-writable" already, like in our case.

invenio_files_rest/config.py
max-moser: An alternative to this new configuration variable altogether would be to have the shared task accept an argument:
```
@shared_task(ignore_result=True)
def clear_orphaned_files(force_delete_orphan_func):
    # ...
```

You could provide the argument (boolean value or function) via the configuration of scheduled celery tasks (`CELERY_BEAT_SCHEDULE`).
Something like:
```
    'clear-orphan-files': {
        'task': 'invenio_files_rest.tasks.clear_orphaned_files',
        'schedule': 60 * 60 * 24,
        'args': (True,)
    },
```

WDYT?

invenio_files_rest/tasks.py
max-moser: Good point, I haven't taken that into consideration (because that's likely not an issue for us any time soon).
Maybe we should add a (configurable) constraint for the maximum number of handled files per execution, with a default value of about 1000 or so?

The only case in which this wouldn't necessarily "work" would be if more orphan files are added between executions than the configured limit... But in that case, one could just increase the limit or schedule the task more frequently.

max-moser: Oh, and of course we could handle a configured limit of `0` to mean "no limit", as is often done... That gives the option to choose freely.

ntarocco: What about having another param, with a default value, that you can configure when you add the celery task in the config as above?
```
'clear-orphan-files': {
  'task': 'invenio_files_rest.tasks.clear_orphaned_files',
  'schedule': 60 * 60 * 24,
  'kwargs': {
    'force': True,
    'limit': 100
  }
},
```
and then:
```suggestion
    orphans = FileInstance.query.filter(~FileInstance.objects.any()).limit(limit).all()
```
?

max-moser: that's pretty much what i was thinking :)
i'll update the PR soon!

max-moser:
It seems that a deprecation was turned into a removal in sphinx 6.
The deprecation warning from sphinx 5:
```
/home/mmoser/.local/share/virtualenvs/invenio-files-rest-bsTvguby/lib/python3.10/site-packages/sphinx/domains/python.py:512: RemovedInSphinx60Warning: Python directive method get_signature_prefix() returning a string is deprecated. It must now return a list of nodes. Return value was '(task)'.
```

PR #282, opened 2022-06-09
migrate to black

utnapischtim:
- migrate to use black as opinionated auto formater
- add .git-blame-ignore-revs
- fix docs compatibilty problem with Sphinx>=5.0.0
- move check_manifest configuration to setup.cfg.
- increase minimal python version to 3.7
- global: clean test infrastructure


tests/test_views_objectversion.py
```diff
@@ -172,169 +183,186 @@ def test_get_unreadable_file(client, headers, bucket, objects, db,
...
-
+        "auth",
+        "objects",
+        "bucket",
+        "location",
+    ],
+)
+def test_get_versions_invalid(client, headers, bucket, objects, permissions, user):
+    """Test object version getting."""
     versions = [
```
max-moser, line 241: could this also be replaced by `parametrize`?
this would make its use a bit more consistent

tests/test_views_bucket.py
```diff
@@ -8,44 +8,50 @@
...
 from flask import url_for
 from testutils import login_user
 
 from invenio_files_rest.models import ObjectVersion
 
 
-def test_head(client, headers, bucket, permissions):
-    """Test checking existence of bucket."""
-    cases = [
+@pytest.mark.parametrize(
```
max-moser, line 18: replaced the for loop over `cases` in the function with parametrization; this is done a few more times in this file

.git-blame-ignore-revs
```diff
@@ -0,0 +1 @@
+01addd4d124b739eb4a560f841cd4e798aa1e36d
```
max-moser, line 1: for ignoring reformatting commits in `git blame` - is this file name the default, or would i need to specify it as CLI argument (or in the config)?

.editorconfig
```diff
@@ -16,17 +16,6 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 charset = utf-8
 
-# Python files
```
max-moser, line 19: was this incompatible with `black`?

run-tests.sh
```diff
@@ -13,17 +14,6 @@ set -o errexit
 # Quit on unbound symbols
 set -o nounset
 
-# Always bring down docker services
```
max-moser, line 16: were the docker services unnecessary?

setup.cfg
```diff
@@ -109,5 +111,14 @@ output-dir = invenio_files_rest/translations/
...
-[pycodestyle]
-ignore = E501
+[isort]
+profile=black
+
+[check-manifest]
+ignore =
+    *-requirements.txt
+
+[tool:pytest]
```
max-moser, line 121: the `pytest.ini` was moved here

max-moser (approved):
just the way i like it: the code is black, the tests are green

max-moser:
i left a few comments, which serve as either questions to @utnapischtim or markers for other reviewers so they don't have to wade through all the formatting changes, because this PR is quite big

to summarize:
* code was formatted with `black`
* a few configs were changed to interfere less with `black`
* code was made compatible with `flask-login` 0.5+ (`user_id` -> `_user_id`)
* some test functions got `parametrize`d (not only because it's fancy but also because it wasn't easily possible to change the logged-in user within the function)
* the test matrix was updated

anything missing?

.git-blame-ignore-revs
utnapischtim: You have nothing to do neither locally nor on github. This change is also existing on all other modules which were migrated to use black.

run-tests.sh
utnapischtim: As you see the tests are running through. It would be really weird if the tests were green but the services were necessary.

tests/test_views_objectversion.py
utnapischtim: I've thought about it, but i don't know how to use the parametrize feature as a double for loop.

run-tests.sh
max-moser: good point! :smile:

tests/test_views_objectversion.py
max-moser: fair enough

run-tests.sh
kpsherva: this will have an impact locally though, it will require manual run of docker container before starting the tests, which is not consistent with other modules

setup.cfg
```diff
@@ -23,18 +24,19 @@ classifiers =
...
     Flask-WTF>=0.15.1
     fs>=2.0.10,<3.0
-    invenio-accounts>=1.4.10
 
 [options.extras_require]
 tests =
+    pytest-black>=0.3.0
     future>=0.18.2
     invenio-access>=1.4.0
+    invenio-accounts>=1.4.10
```
kpsherva, line 39: why is it necessary to add the new dependency ?

run-tests.sh
utnapischtim: If that would have an impact locally the tests on github would also not work, but they are passing. For me that shows that the running docker containers are not necessary for the tests of this package. Further i tried another time locally to be sure that no containers are up and running. It worked locally without running containers.

setup.cfg
utnapischtim: You are right. It is not necessary. It will be installed anyway. My thinking was to explicitly show which packages are used in the tests.

kpsherva: the problem with that approach is that we would have to maintain the minimal version in several places, instead of one

utnapischtim: removed

run-tests.sh
utnapischtim: readded

PR #283, opened 2022-06-30
send file: add params for cache control headers

NRodriguezcuellar:
added 2 params called max_age and no_cache which will allow more control about what cache control headers are being returned by the file response object

partly closes https://github.com/inveniosoftware/invenio-communities/issues/718


invenio_files_rest/helpers.py
```diff
@@ -166,11 +181,14 @@ def send_stream(stream, filename, size, mtime, mimetype=None, restricted=True,
     # Set cache-control
     if not restricted:
         rv.cache_control.public = True
-        cache_timeout = current_app.get_send_file_max_age(filename)
+        cache_timeout = current_app.get_send_file_max_age(filename) or max_age
         if cache_timeout is not None:
             rv.cache_control.max_age = cache_timeout
             rv.expires = int(time() + cache_timeout)
 
+    if no_cache:
```
ntarocco, line 189: I would not add an extra param: if `restricted`, it should be `no-cache`. WDYT?
```suggestion
    else:
```

invenio_files_rest/helpers.py
```diff
@@ -166,11 +181,14 @@ def send_stream(stream, filename, size, mtime, mimetype=None, restricted=True,
     # Set cache-control
     if not restricted:
         rv.cache_control.public = True
-        cache_timeout = current_app.get_send_file_max_age(filename)
+        cache_timeout = current_app.get_send_file_max_age(filename) or max_age
```
ntarocco, line 184: Why the func is not enough?

NRodriguezcuellar: because that is a globally set default, i want to make it changeable also per function call, as is possible with the flask `send_file` function

NRodriguezcuellar: I just realized it should be the other way around then :smile:

invenio_files_rest/helpers.py
NRodriguezcuellar: could be, that would basically make it the default because unless restricted is changed it is `True`.

invenio_files_rest/helpers.py
```diff
@@ -166,11 +176,17 @@ def send_stream(stream, filename, size, mtime, mimetype=None, restricted=True,
...
+        cache_timeout = (
+            current_app.get_send_file_max_age(filename)
+            or current_app.config["FILES_REST_SEND_FILE_MAX_AGE"]
+        )
         if cache_timeout is not None:
             rv.cache_control.max_age = cache_timeout
             rv.expires = int(time() + cache_timeout)
 
+    if restricted:
+        rv.cache_control.no_cache = True
```
ntarocco, line 188: @slint can you please check if this change and the one above (the new var `current_app.config["FILES_REST_SEND_FILE_MAX_AGE"]`) are ok with you?

ntarocco (commented):
LGTM! Better to have also @slint to check this

invenio_files_rest/helpers.py
slint: - Regarding, `FILES_REST_SEND_FILE_MAX_AGE`, can't we just document/comment that setting  [`SEND_FILE_MAX_AGE_DEFAULT`](https://flask.palletsprojects.com/en/2.1.x/config/#SEND_FILE_MAX_AGE_DEFAULT), which controlls `current_app.get_send_file_max_age(...)` should be enough?
- For the restricted files, I agree with the `no-cache`

NRodriguezcuellar: We could set the default in the cookiecutter `invenio.cfg` too. For any instances made with older cookiecutters they might still have an undefined max_age though.  WDTY? @ntarocco

ntarocco: It should be ok because `SEND_FILE_MAX_AGE_DEFAULT` is part of Flask framework, added long time ago...

NRodriguezcuellar: So we should not set it in `invenio.cfg` ?

ntarocco:
LGTM! Tests failing because of `FAILED invenio_files_rest/helpers.py::PYCODESTYLE`.

invenio_files_rest/helpers.py
```diff
@@ -166,11 +176,17 @@ def send_stream(stream, filename, size, mtime, mimetype=None, restricted=True,
...
         rv.cache_control.public = True
+        # get the value of the flask config variable "SEND_FILE_MAX_AGE_DEFAULT" and use it to set how long (in
+        # seconds) a file can be cached before it is considered stale and the client should fetch the newest version.
         cache_timeout = current_app.get_send_file_max_age(filename)
+
         if cache_timeout is not None:
             rv.cache_control.max_age = cache_timeout
             rv.expires = int(time() + cache_timeout)
 
+    if restricted:
```
ntarocco, line 187: Since you are changing the code because tests are failing, just add an `else` here?
```suggestion
    else:
```

invenio_files_rest/helpers.py
```diff
@@ -166,10 +176,15 @@ def send_stream(stream, filename, size, mtime, mimetype=None, restricted=True,
     # Set cache-control
     if not restricted:
         rv.cache_control.public = True
+        # get the value of the flask config variable "SEND_FILE_MAX_AGE_DEFAULT" and use it to set how long (in
+        # seconds) a file can be cached before it is considered stale and the client should fetch the newest version.
```
slint, line 180: I would go a bit softer with the details, since these might change also in the future :)

```suggestion
        # See flask config variable "SEND_FILE_MAX_AGE_DEFAULT"
        # https://flask.palletsprojects.com/en/2.1.x/api/#flask.Flask.get_send_file_max_age
```

slint (commented):
LGTM!

One less important but good thing to clarify and align on, given the recent formatting efforts across Invenio repos (cc @kpsherva): Would it be possible to separate the actual change in this PR (i.e. the caching behaviour) from the Black formatting in order to keep the history clean? Just separating the commits might be enough since they can be added into the `.git-blame-ignore-revisions` file later.

invenio_files_rest/helpers.py
slint: In that case, I trust that Flask docs keep their links stable and explain in more detail (or point to other resources), vs. us maintaining details of how browser-side caching works. If this breaks in the future, of course we'll have to fix anyways, but I wouldn't also add into it maintaining a comment regarding how caching works, but delegate to [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age)

NRodriguezcuellar: that is fair 😄 , made the change.

Pineirin:
LGTM, but can you fix the tests?

constraints-pypi.txt
```diff
@@ -1 +1,2 @@
 # Empty on purpose
+flask-login==0.4.1
```
kpsherva, line 2: could you remove the pin? it is not needed anymore

PR #298, opened 2023-11-01
models: add remove-many

jrcastro2:
* Closes https://github.com/zenodo/rdm-project/issues/413

### DIFF
--- a/invenio_files_rest/models.py
+++ b/invenio_files_rest/models.py
@@ -1307,6 +1307,25 @@ class ObjectVersion(db.Model, Timestamp):
         return None
 
     @classmethod
+    def remove_many(cls, version_ids, bucket):
+        """Permanently removes specific object versions from the database.
+
+        :param version_ids: list of version_id.
+        :param bucket: The bucket (instance or id) to delete the object from.
+        """
+        bucket = as_bucket(bucket)
+        if bucket.locked:
+            raise BucketLockedError()
+        object_versions = cls.query.filter(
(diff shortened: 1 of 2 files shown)


invenio_files_rest/models.py
```diff
@@ -1306,6 +1306,25 @@ def delete(cls, bucket, key):
...
+        """
+        bucket = as_bucket(bucket)
+        if bucket.locked:
+            raise BucketLockedError()
+        object_versions = cls.query.filter(
+            cls.version_id.in_(version_ids), cls.bucket_id == bucket.id
+        )
+        for object_version in object_versions.all():
+            bucket.size -= object_version.file.size
+        object_versions.delete()
```
kpsherva, line 1324: judging from the delete function above, shouldn't we do the same? ie. create an empty object version instead of deleting ? I think this might be hard delete, removing all the revisions (I might be wrong)

invenio_files_rest/models.py
```diff
@@ -1306,6 +1306,25 @@ def delete(cls, bucket, key):
             return cls.create(bucket, key)
         return None
 
+    @classmethod
+    def remove_many(cls, version_ids, bucket):
```
kpsherva, line 1310: minor: 
```suggestion
    def delete_many(cls, version_ids, bucket):
```

jrcastro2: There are 2, remove and delete. Remove is hard delete (the one causing the slowness) while delete is soft delete. So far we only need the multiple delete for hard deletion that's why I name it remove_many.

slint (approved):

PR #299, opened 2023-11-03
Add autoflush to get file

jrcastro2:
* closes https://github.com/zenodo/rdm-project/issues/508

### DIFF
--- a/invenio_files_rest/models.py
+++ b/invenio_files_rest/models.py
@@ -36,12 +36,14 @@ model.
 import re
 import sys
 import uuid
+from copy import deepcopy
 from datetime import datetime
 from functools import wraps
 from os.path import basename
 
 from flask import current_app
 from invenio_db import db
+from sqlalchemy import insert, inspect
 from sqlalchemy.dialects import mysql
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import validates
@@ -584,8 +586,10 @@ class Bucket(db.Model, Timestamp):
         :returns: ``self``.
         """
         with db.session.begin_nested():
-            ObjectVersion.query.filter_by(bucket_id=self.id).delete()
-            self.query.filter_by(id=self.id).delete()
+            ObjectVersion.query.filter_by(bucket_id=self.id).delete(
+                synchronize_session=False
+            )
+            self.query.filter_by(id=self.id).delete(synchronize_session=False)
         return self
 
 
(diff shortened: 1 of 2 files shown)


invenio_files_rest/models.py
```diff
@@ -719,7 +721,8 @@ def validate_uri(self, key, uri):
     @classmethod
     def get(cls, file_id):
         """Get a file instance."""
-        return cls.query.filter_by(id=file_id).one_or_none()
+        with db.session.no_autoflush:
```
kpsherva, line 724: this change to be tested - not sure of the consequence

invenio_files_rest/models.py
```diff
@@ -1161,6 +1164,37 @@ def copy(self, bucket=None, key=None):
 
         return new_ob
 
+    def clone(self):
+        db.session.expunge(self)
+        make_transient(self)
```
kpsherva, line 1169: done via python objects due to sqlalchemy not being flexible enough with insert values and instert from select 
another alternative would be doing it via pure sql queries but it does not really follow best practices

tests/test_models.py
```diff
@@ -735,6 +735,41 @@ def test_object_copy(app, db, dummy_location):
...
+
+    # TODO: add test for deleted object_versions
+
+    # Copy
+    key1 = "selftest1"
+    key2 = "selftest2"
+    obj1 = ObjectVersion.create(b1, key1).set_file(f1)
+    obj2 = ObjectVersion.create(b1, key2).set_file(f2)
+    db.session.commit()
+    object_versions = ObjectVersion.copy_many([obj1.version_id, obj2.version_id], b2)
```
kpsherva, line 756:
```suggestion
    object_versions = ObjectVersion.copy_from(b2)
```

invenio_files_rest/models.py
```diff
@@ -1161,6 +1165,45 @@ def copy(self, bucket=None, key=None):
...
+            for o in ObjectVersion.get_by_bucket(bucket_id):
+                new_ov = deepcopy(
+                    {
+                        key: value
+                        for key, value in o.__dict__.items()
+                        if key in obj_columns
+                    }
+                )
+                new_ov["bucket_id"] = destination_bucket_id
+                new_ov["version_id"] = uuid.uuid4()
```
kpsherva, line 1199: isn't uuid created on insert? or it does not work like that for bulk? just curious

invenio_files_rest/models.py
```diff
@@ -1161,6 +1165,45 @@ def copy(self, bucket=None, key=None):
...
+                new_versions.append(new_ov)
+
+            if new_versions:
+                db.session.execute(insert(ObjectVersion), new_versions)
```
kpsherva, line 1203: do we need object version tags? I think if you don't add them then you cannot upgrade cds-videos, but I might be wrong

zzacharo: In cds-videos we don't use the `copy_from` method. What do you mean by not able to upgrade? Maybe I miss something...

slint (approved):

invenio_files_rest/models.py
slint: I get what @kpsherva means here, but since there's no plan to use this method (besides in InvenioRDM), I wouldn't go down the rabbit hole of also implementing copying tags of each object version.

I'll shelve into a separate issue.

PR #301, opened 2023-11-10
utils: add fallback mimetype

jrcastro2:
* closes https://github.com/zenodo/rdm-project/issues/547

### DIFF
--- a/invenio_files_rest/utils.py
+++ b/invenio_files_rest/utils.py
@@ -54,4 +54,4 @@ def guess_mimetype(filename):
     m, encoding = mimetypes.guess_type(filename)
     if encoding:
         m = ENCODING_MIMETYPES.get(encoding, None)
-    return m or "application/octet-stream"
+    return m or current_app.config["FILES_REST_FALLBACK_MIMETYPE"]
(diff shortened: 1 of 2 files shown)


invenio_files_rest/utils.py
```diff
@@ -21,6 +21,9 @@
 }
 """Mapping encoding to MIME types which are not in mimetypes.types_map."""
 
+FALLBACK_MIMETYPE = "application/octet-stream"
```
kpsherva, line 24: does it make sense to expose it as config ? ping @slint

slint: Actually, this doesn't make sense to have as config, the fallback for "I've no idea what I'm serving" is `application/octet-stream` (i.e. "you get bytes").

jrcastro2:
Not required anymore as we agreed to hardcode the default mimetype

slint:
Just to clarify, I'm against moving this to config since IMHO it's a constant, and I also agree that introducing it in `utils.py` as `FALLBACK_MIMETYPE` might have been a "brittle" solution.

PR #304, opened 2024-01-17
models: adds index on version_id and bucket_id

jrcastro2:
* adds alembic recipe
* closes zenodo/rdm-project#640


slint:
Apparently MySQL anyways automatically creates indices on FKs and dropping the index [is not allowed](https://stackoverflow.com/questions/8482346/mysql-cannot-drop-index-needed-in-a-foreign-key-constraint).

@jrcastro2 can you look into if it's possible to conditionally in the downgrade recipe to not drop the index if we're on MySQL?   If this isn't easy then we can remove `mysql` from the test matrix and open an issue.

invenio_files_rest/alembic/e172c837b036_add_indexes.py
```diff
@@ -0,0 +1,64 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Add indexes."""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "e172c837b036"
+down_revision = "a29271fd78f8"
```
jrcastro2, line 15: Another alembic recipe was merged after creating this one, so we should update this down_revision to point to that one to keep the correct flow.

slint (approved):
I'll

PR #309, opened 2024-09-03
fix: fix downloading for non-latin-1 filenames containing a comma

martinobersteiner:
### Description

one of our invenio-users at TU Graz uploaded a file named *Erläuterungen zum Analysebogen digitaler Schulbücher 1-2 (Design, Technik, Barrierefreiheit)* ([link to record-page](https://repository.tugraz.at/oer/yzhp8-7ay47))
attempting to download this file from Google Chrome fails with `net::ERR_RESPONSE_HEADERS_MULTIPLE_CONTENT_DISPOSITION`

the same issue occurs for a vanilla invenio-installation:
download_weird-filenames

this download-failure occurs for the following filenames (seemingly independent of file-content):
- has a file-extension for which a *Content-Disposition* header is generated, e.g. `.docx`
- contains a comma (`,`) character
- is NOT `latin-1`-encodable

simplest example I could find is `ä, B.docx` (file-contents are an empty `.docx` file)
note that `ä`'s unicode-codepoints here are `\U+0061\U+0308` (i.e. `a` followed by *Combining Diaeresis*), which isn't `latin-1`-encodable
(unicode `\U+00E4` (`ä` *Latin Small Letter A with Diaeresis*) is `latin-1`-encodable)
`ä, B.docx` has this *Content-Disposition* header: `attachment; filename*=UTF-8''a%CC%88,%20B.docx; filename="a, B.docx"`

the issue lies in [invenio_files.rest/helpers.py:L150](https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/helpers.py#L150):
the there-used `werkzeug.urls.url_quote` does not quote comma (`,`) characters
hence the *Content-Disposition* header (`... filename*=UTF-8''a%CC%88,%20B.docx ...`) contains an unquoted `,`, which makes Google Chrome interpret the header as two filenames (`a%CC%88` and `%20B.docx`)
but *Content-Disposition* headers may only contain one filename, causing the above-mentioned `net::ERR_RESPONSE_HEADERS_MULTIPLE_CONTENT_DISPOSITION`

solution:
`werkzeug.urls.url_quote` has been deprecated by `werkzeug` anyway
in [their deprecation-notice](https://github.com/pallets/werkzeug/pull/2768) they suggest using std-lib's `urllib.parse.quote` instead
(*"Remove our copy of `urllib.parse` from our `urls` module, in favor of using the built-in module directly."*)
`urllib.parse.quote` correctly quotes `,` (to `%C2`), solving the issue

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [x] I've added relevant documentation.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_files_rest/helpers.py
+++ b/invenio_files_rest/helpers.py
@@ -14,11 +14,10 @@ import os
 import unicodedata
 import warnings
 from time import time
-from urllib.parse import urlsplit
+from urllib.parse import quote, urlsplit
 
 from flask import current_app, make_response, request
 from werkzeug.datastructures import Headers
-from werkzeug.urls import url_quote
 from werkzeug.wsgi import FileWrapper
 
 MIMETYPE_TEXTFILES = {"readme"}
@@ -147,7 +146,10 @@ def send_stream(
         try:
             filenames = {"filename": filename.encode("latin-1")}
         except UnicodeEncodeError:
-            filenames = {"filename*": "UTF-8''%s" % url_quote(filename)}
+            # in the header-field Content-Disposition, filename* is a "extended parameter" (RFC 8187)
+            # safe=characters that need not be %-encoded in extended parameter's value (RFC 8187 `attr-char`)
+            quoted = quote(filename, safe="!#$%&+-.^_`|~")
+            filenames = {"filename*": f"UTF-8''{quoted}"}
             encoded_filename = unicodedata.normalize("NFKD", filename).encode(
                 "latin-1", "ignore"
             )
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,7 @@
 #
 # This file is part of Invenio.
 # Copyright (C) 2020 CERN.
-# Copyright (C) 2022 Graz University of Technology.
+# Copyright (C) 2022, 2024 Graz University of Technology.
 #
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -25,76 +25,7 @@ on:
         default: 'Manual trigger'
 
 jobs:
-  Tests:
-    runs-on: ubuntu-20.04
-    timeout-minutes: 20
-    strategy:
-      matrix:
-          python-version: [3.7, 3.8, 3.9]
-          requirements-level: [pypi]
-          cache-service: [redis]
-          db-service: [postgresql13, postgresql14, mysql5, mysql8]
-          exclude:
-          - python-version: 3.7
-            db-service: postgresql14
-
-          - python-version: 3.8
-            db-service: mysql8
-
-          - python-version: 3.8
-            db-service: mysql5
-
-          - python-version: 3.9
-            db-service: mysql5
-
-          include:
-          - db-service: postgresql13
-            DB_EXTRAS: "postgresql"
-
-          - db-service: postgresql14
-            DB_EXTRAS: "postgresql"
-
-          - db-service: mysql5
-            DB_EXTRAS: "mysql"
-
-          - db-service: mysql8
-            DB_EXTRAS: "mysql"
-
-    env:
-      CACHE: ${{ matrix.cache-service }}
-      DB: ${{ matrix.db-service }}
-      EXTRAS: tests
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
-        with:
-          python-version: ${{ matrix.python-version }}
-
-      - name: Generate dependencies
-        run: |
-          pip install wheel requirements-builder
-          requirements-builder -e "$EXTRAS" --level=${{ matrix.requirements-level }} setup.py > .${{ matrix.requirements-level }}-${{ matrix.python-version }}-requirements.txt
-
-      - name: Cache pip
-        uses: actions/cache@v2
-        with:
-          path: ~/.cache/pip
-          key: ${{ runner.os }}-pip-${{ hashFiles('.${{ matrix.requirements-level }}-${{ matrix.python-version }}-requirements.txt') }}
-
-      - name: Install dependencies
-        run: |
-          pip install -r .${{ matrix.requirements-level }}-${{ matrix.python-version }}-requirements.txt -c constraints-${{ matrix.requirements-level }}.txt
-          pip install ".[$EXTRAS]"
-          pip freeze
-          docker --version
-          docker-compose --version
-
-      - name: Run translations test
-        run: ./run-i18n-tests.sh
-
-      - name: Run tests
-        run: |
-          ./run-tests.sh
+  Python:
+    uses: inveniosoftware/workflows/.github/workflows/tests-python.yml@master
+    with:
+      extras: "tests,redis"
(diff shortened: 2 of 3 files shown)


fenekku (approved):
Good investigative work and explanation. I guess the latin-1 encodable part isn't addressed by this but it's good enough, right? Thanks for this!

martinobersteiner:
> Good investigative work and explanation. I guess the latin-1 encodable part isn't addressed by this but it's good enough, right? Thanks for this!

actually, the `latin-1` case may not need a fix:
in the `latin-1` case, quotes (`"`) are put around the header's filename like so `... filename*="quoted-filename,commas-are-fine-here" ...`

in my local testing at least, this allowed `latin-1`-filenames-containing-commas to be downloadable

(didn't test yet-more-weird filenames though; unix-filenames can even have control characters in there, I think?)

kpsherva:
thank you for providing the detailed context for the reviewers, it is greatly appreciated.
suggestion for failing tests fix: https://github.com/inveniosoftware/invenio-files-rest/blob/master/.github/workflows/tests.yml#L93

kpsherva (approved):

martinobersteiner:
I'm assuming the test I fixed meant to check whether `invenio_files_rest.views:merge_multipartobject.delay` was called with the correct upload-id

however, `MagicMock.called_with` never did that: in older cpython versions, `.called_with(...)` only returns another `MagicMock`, without ever asserting anything
newer cpython versions forbid the attribute name `.called_with` as it's a common mis-naming of `.assert_called_with` (see [this commit](https://github.com/python/cpython/commit/1d4d677d1c90fcf4886ded0bf04b8f9d5b60b909))

however, `task.delay.assert_called_with` isn't directly usable either, because `task.delay` is called not only with an `upload-id` arg but also with a kwarg `version_id`
and we don't know the value of `version_id` at this point

that's why I'm checking `call_args.args` here (purposefully ignoring `call_args.kwargs` that contains the unknown `version_id`)

invenio_files_rest/helpers.py
```diff
@@ -147,7 +146,7 @@ def send_stream(
         try:
             filenames = {"filename": filename.encode("latin-1")}
         except UnicodeEncodeError:
-            filenames = {"filename*": "UTF-8''%s" % url_quote(filename)}
+            filenames = {"filename*": "UTF-8''%s" % quote(filename)}
```
kpsherva, line 149: suggestion: it would be good to align with https://github.com/zenodo/zenodo-rdm/blob/8b81afaeba88ba00230bfcf189f6b62b9ef8adb0/site/zenodo_rdm/files.py#L125C46-L125C51
https://www.rfc-editor.org/rfc/rfc5987

slint, line 149:
```suggestion
            filenames = {"filename*": "UTF-8''%s" % quote(filename, safe="!#$&+-.^_`|~")}
```

I see in Zenodo after some post-migration fixes we explicitly [pass a `safe=...` parameter](https://github.com/zenodo/zenodo-rdm/blob/8b81afaeba88ba00230bfcf189f6b62b9ef8adb0/site/zenodo_rdm/files.py#L125). I suspect this list comes from [this werkzeug util](https://github.com/pallets/werkzeug/blob/5add63c955131fd73531d7369f16b2f1b4e342d4/src/werkzeug/utils.py#L462C49-L462C61).

Does it make sense to reuse here, or have a look at the `werkzeug` code to verify we're roughly doing the same handling? Unfortunately, we're pinned to a lower version of werkzeug, so we can just upgrade and reuse the upstream function.

martinobersteiner: thanks for the suggestion
I did align this PR with the code you linked

martinobersteiner: looking into it some more, `filename*` is an *extended parameter* for the header-field *Content-Disposition*
*extended parameter*s for header-fields are defined in [RFC 8187](https://www.rfc-editor.org/rfc/rfc8187)
in its [section 3.2.1](https://www.rfc-editor.org/rfc/rfc8187#section-3.2.1), the exact characters used by `werkzeug` (`"!#$&+-.^_`|~"`) are defined as safe to use without percent-encoding as values for extended fields

so adding `safe="!#$&+-.^_`|~"` makes sense imo

martinobersteiner: I think if at all we should use `flask.send_file` instead of `werkzeug.utils.send_file`

If I understand correctly, there's a reason that `invenio`'s function is called `send_stream` instead of `send_file` though:
we're working with an abstract file-system provided by `pyfs`, so the only thing we're guaranteed to have is a *stream*, not necessary a *file* (e.g. could be entry in a zip-file, abstracted by `pyfs` to behave like a filesystem)

now, streams can be passed to `flask.send_file` and work just fine in my local testing
*but*, we'd lose *Content-Length* and *Last-Modified* headers
(`flask.send_file` tries to get those headers from the passed file, but leaves them out when passed a stream)
(`invenio_files_rest.helpers.send_stream` has additional parameters for these headers instead)

not sure how easy those could be added if we were to use `flask.send_file`...

martinobersteiner: Comparing our current code with `werkzeug 3.0`, there's some differences wrt *Content-Disposition* header:
- `werkzeug` adds `filename=...` and `filename*=..` to *Content-Disposition* whether it's `inline` or `attachment`, whereas we just add it in the `attachment` case
- `werkzeug`'s `filename` is `ascii` whereas ours is `latin-1`

*Content-Disposition* header is defined in [RFC 62266](https://www.rfc-editor.org/rfc/rfc6266)
in [its *Appendix D*](https://www.rfc-editor.org/rfc/rfc6266#appendix-D), it lists some advice for this header
reading that, I think `werkzeug` is the more correct one in both of the cases I listed above...

there might be further differences between `invenio_files_rest.helpers.stream_file` and `flask.send_file` for other headers...

PR #312, opened 2024-09-26
objecttag: Make the max length of the tag values configurable (defaul…

psaiz:
…t to 255)

:heart: Thank you for your contribution!

### Description
For Open Data, we store the 'cold' url of files as an object tag. Some of the uri are longer than 255, so it would be nice to set the limit as a parameter that can be tuned by the different instances.

### DIFF
--- a/invenio_files_rest/config.py
+++ b/invenio_files_rest/config.py
@@ -68,6 +68,14 @@ FILES_REST_OBJECT_KEY_MAX_LEN = 255
    with PostgreSQL database.
 """
 
+FILES_REST_OBJECT_TAG_VALUE_MAX_LEN = 255
+"""Maximum length of the ObjectTag.value.uri field.
+
+.. warning::
+   Setting this variable to anything higher than 255 is only supported
+   with PostgreSQL database.
+"""
+
 FILES_REST_FILE_URI_MAX_LEN = 255
 """Maximum length of the FileInstance.uri field.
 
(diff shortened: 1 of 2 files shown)


slint (approved):

utnapischtim:
which `uri` column is used to store the configurable value?

- https://github.com/inveniosoftware/invenio-files-rest/blob/400a34f2ca515426e7226acb390a63b438810b44/invenio_files_rest/models.py#L273
- https://github.com/inveniosoftware/invenio-files-rest/blob/400a34f2ca515426e7226acb390a63b438810b44/invenio_files_rest/models.py#L686

i am asking because i think only the second can store an character string greater then 255 characters

psaiz:
Thanks for checking, @utnapischtim. This is to store on the tags, so it should be https://github.com/inveniosoftware/invenio-files-rest/blob/400a34f2ca515426e7226acb390a63b438810b44/invenio_files_rest/models.py#L1449 (which is already text)

invenio_files_rest/config.py
```diff
@@ -68,6 +68,14 @@
    with PostgreSQL database.
 """
 
+FILES_REST_OBJECT_TAG_VALUE_MAX_LEN = 255
+"""Maximum length of the ObjectTag.value.uri field.
```
slint, line 72:
```suggestion
"""Maximum length of the ObjectTag.value field.
```

minor: I know you store `uri`s there, but not everyone does ;)

PR #315, opened 2024-11-28
bump major

utnapischtim:
- setup: bump invenio dependencies
- release: v3.0.0

### DIFF
--- a/invenio_files_rest/alembic/8ae99b034410_create_files_objecttags_table.py
+++ b/invenio_files_rest/alembic/8ae99b034410_create_files_objecttags_table.py
@@ -2,6 +2,7 @@
 #
 # This file is part of Invenio.
 # Copyright (C) 2017-2019 CERN.
+# Copyright (C) 2024 Graz University of Technology.
 #
 # Invenio is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -43,7 +44,9 @@ def upgrade():
         "files_objecttags",
         sa.Column("version_id", sqlalchemy_utils.types.uuid.UUIDType(), nullable=False),
         sa.Column(
-            "key", sa.Text().with_variant(mysql.VARCHAR(255), "mysql"), nullable=False
+            "key",
+            sa.String(255).with_variant(mysql.VARCHAR(255), "mysql"),
+            nullable=False,
         ),
         sa.Column(
             "value", sa.Text().with_variant(mysql.VARCHAR(255), "mysql"), nullable=False
--- a/invenio_files_rest/helpers.py
+++ b/invenio_files_rest/helpers.py
@@ -143,19 +143,18 @@ def send_stream(
     # Force Content-Disposition for application/octet-stream to prevent
     # Content-Type sniffing.
     if as_attachment or mimetype == "application/octet-stream":
-        # See https://github.com/pallets/flask/commit/0049922f2e690a6d
+        # see https://github.com/pallets/werkzeug/blob/main/src/werkzeug/utils.py#L456-L465
         try:
-            filenames = {"filename": filename.encode("latin-1").decode("utf-8")}
+            filename.encode("ascii")
         except UnicodeEncodeError:
-            # in the header-field Content-Disposition, filename* is a "extended parameter" (RFC 8187)
-            # safe=characters that need not be %-encoded in extended parameter's value (RFC 8187 `attr-char`)
-            quoted = quote(filename, safe="!#$%&+-.^_`|~")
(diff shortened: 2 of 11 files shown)


invenio_files_rest/alembic/8ae99b034410_create_files_objecttags_table.py
```diff
@@ -43,7 +44,9 @@ def upgrade():
         "files_objecttags",
         sa.Column("version_id", sqlalchemy_utils.types.uuid.UUIDType(), nullable=False),
         sa.Column(
-            "key", sa.Text().with_variant(mysql.VARCHAR(255), "mysql"), nullable=False
+            "key",
+            sa.String(255).with_variant(mysql.VARCHAR(255), "mysql"),
```
slint, line 48:
```suggestion
            sa.Text().with_variant(mysql.VARCHAR(255), "mysql"),
```

I think this has to stay as `sa.Text()` since there's an expectation that in PostgreSQL you can have a string as long as one needs.

invenio_files_rest/helpers.py
```diff
@@ -144,7 +145,7 @@ def send_stream(
     if as_attachment or mimetype == "application/octet-stream":
         # See https://github.com/pallets/flask/commit/0049922f2e690a6d
         try:
-            filenames = {"filename": filename.encode("latin-1")}
+            filenames = {"filename": filename.encode("latin-1").decode("utf-8")}
```
slint, line 148: I feel this is fine, since I remember that `werkzeug` started dealing with encoding/decoding these values a bit better in later versions (which is the reason we've copy-pasted some of this logic here).

It might be even better to remove the logic entirely... but this needs some testing probably

tests/test_request.py
```diff
@@ -30,7 +31,7 @@ def test():
 
     with app.test_client() as client:
         # No content-type, no max content length checking
-        data = b"a" * (max_len + 1)
+        data = b"a" * (max_len - 1)
```
slint, line 34: This one is strange... Maybe the test was bad/faulty from before? Or did the test client change maybe, and now the `conten-type` is passed implicitly?

invenio_files_rest/alembic/8ae99b034410_create_files_objecttags_table.py
utnapischtim: the problem is that the type in the alembic recipe and the type in the model are not the same.

model has db.String(255) [here](https://github.com/inveniosoftware/invenio-files-rest/blob/8e09e6152c6b195ce439f6a665624a2ddd274dd5/invenio_files_rest/models.py#L1461)
alembic recipe has db.TEXT [here](https://github.com/inveniosoftware/invenio-files-rest/blob/master/invenio_files_rest/alembic/8ae99b034410_create_files_objecttags_table.py#L46)

older alembic or sqlalchemy didn't see a problem with that, but now we have one

tests/test_request.py
utnapischtim: as i understand the [PR](https://github.com/pallets/werkzeug/commit/4f7048e7a31752142f18eefeccd49acc42a89e31#diff-f94258ec650ac61110aed42f4708d43590a6a3e8ebb76234d42a933c3796886d) the thing is that the max_content_length hasn't been checked before.

now it is checked.

invenio_files_rest/helpers.py
utnapischtim: rethinking the encode/decode part should be done, but not know. this would need some further testing and would delay the process of moving to flask>=3.0.

utnapischtim: i could use [this](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/utils.py#L456-L465) instead of what we have now

invenio_files_rest/alembic/8ae99b034410_create_files_objecttags_table.py
slint: Yikes, then yes that was a mistake, so this fix is valid :)

invenio_files_rest/helpers.py
utnapischtim: would [that](https://github.com/inveniosoftware/invenio-files-rest/pull/315/commits/d0a6fe692dfbcdaf7e69298ee74b32bdc16faa35) what you had in mind?

slint: Yes, that would work as well. I'm still a tiny bit curious why we have to do this on our side, instead of just delegating to Flask/werkzeug (but that can be addressed separately)

slint (approved):

PR #321, opened 2025-04-23
Adds range request support for accessing files

spilth:
### Description

This PR adds support for HTTP Range Request when accessing files. This is important for file types like [WACZ](https://www.loc.gov/preservation/digital/formats/fdd/fdd000586.shtml) or [PMTiles](https://docs.protomaps.com/pmtiles/) which can be quite large and are best accessed piecemeal.

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [x] I've added relevant documentation.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).

**Reminder**

By using GitHub, you have already agreed to the [GitHub’s Terms of Service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license) including that:

1. You license your contribution under the same terms as the current repository’s license.
2. You agree that you have the right to license your contribution under the current repository’s license.

### DIFF
--- a/invenio_files_rest/helpers.py
+++ b/invenio_files_rest/helpers.py
@@ -14,12 +14,15 @@ import mimetypes
 import os
 import unicodedata
 import warnings
+from http import HTTPStatus
 from time import time
 from urllib.parse import quote, urlsplit
 
 from flask import current_app, make_response, request
 from werkzeug.datastructures import Headers
 from werkzeug.wsgi import FileWrapper
+
+from .wsgi import RangeFileWrapper
 
 MIMETYPE_TEXTFILES = {"readme"}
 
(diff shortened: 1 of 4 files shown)


invenio_files_rest/helpers.py
```diff
@@ -119,7 +122,58 @@ def send_stream(
...
     # Construct headers
     headers = Headers()
-    headers["Content-Length"] = size
+
+    range_header = request.headers.get("Range")
+    partial_response = False
+    start_byte = 0
+    end_byte = size - 1
+
+    if (
```
slint, line 131: minor: is there a way that we avoid doing all this parsing and delegate it to werkzeug? I see via [`Request..make_conditional(...)`](https://github.com/pallets/werkzeug/blob/7868bef5d978093a8baa0784464ebe5d775ae92a/src/werkzeug/wrappers/response.py#L758) and [in `._process_range_request()`](https://github.com/pallets/werkzeug/blob/7868bef5d978093a8baa0784464ebe5d775ae92a/src/werkzeug/wrappers/response.py#L654), there's a big part of the logic that handles parsing the `Range` header, setting correct response headers, and [wrapping in the response in the `_RangeFileWrapper`](https://github.com/pallets/werkzeug/blob/7868bef5d978093a8baa0784464ebe5d775ae92a/src/werkzeug/wrappers/response.py#L702) that was copied here.

I only see some potential roadblocks in terms of taking into account the new  `FILES_REST_ALLOW_RANGE_REQUESTS` config to control this, but otherwise I feel it would simplify the code a lot.

slint (commented):
Thanks a lot for this @spilth, it looks pretty much complete and good to go :)

I added a single comment about possibly simplifying this even further, being a bit concerned about copying code from werkzeug (maintainability-wise). If you tried and this was unavoidable, my only preference would be to move the big `if ...` block in a separate function for readability.

invenio_files_rest/helpers.py
spilth: I'll take a look and see if that's possible. I was trying to honor their private APIs since that can be a slippery slope support-wise.

spilth: Okay, thank you for pointing me in that direction. I've made the update.

The only downside is that invalid requests now return a 500 instead of 416, due to how werkzeug works.

tmorrell (approved):
This looks like a good improvement and has complete tests. We should try to get it into v13.

slint (approved):

PR #335, opened 2025-07-27
fix: utcnow deprecation warning

utnapischtim:


utnapischtim:
closed in favor of #336 

if the problems should be handled independently, please update this PR

